Issue #21, opened 2017-05-13
Fix Babel/polyfills


brianmhunt:
To be backwards compatible the TKO library needs to be polyfilled.  Examples of ES6 features that need polyfill:

- Generators
- Object.entries/Object.assign
- Backticks
- `=>` functions
- Spreads

Everything in TKO should be polyfillable (i.e. no WeakSet), but my attempts to make it work, using the config below (among many variants), never worked quite right.

An example of the Babel attempt, removed in 7dbec4cab30259e87c5c9f03f7a8137b1c9c8057, is:

- [`package.json`](https://github.com/knockout/tko/blob/7dbec4cab30259e87c5c9f03f7a8137b1c9c8057/package.json)
- [`.babelrc`](https://github.com/knockout/tko/blob/baf5a4033e28d0e3c2ca2ae83f7caa7f9b6ed6ff/.babelrc)
- [`karma.conf.js`](https://github.com/knockout/tko/blob/7dbec4cab30259e87c5c9f03f7a8137b1c9c8057/karma.conf.js)
- [`rollup.config.js`](https://github.com/knockout/tko/blob/7dbec4cab30259e87c5c9f03f7a8137b1c9c8057/rollup.config.js)

caseyWebb:
I'm interested in helping out with this one.

If I may, I'd like to suggest using Typescript instead of Babel for transpilation. At this point, there is very little that babel can do that tsc can't, and it's faster. If necessary, it's trivial to add babel to cover what TypeScript can't.

The coolest part about TS in my opinion is the incremental adoption. The build process can utilize the tsc without any further buy-in. Also has the upside of making publishing typings trivial.

brianmhunt:
TS is fine.  I have no idea how it works, but so long as it does, I'm good with that. :)

codymullins:
I'd +1 the Typescript too.

brianmhunt:
@caseyWebb  I've gotten testing working, but I had to make some changes to the packages, per  67ec170b6a2d573b6ee013474d3ef38cee90d087 and c07a2d192beebff07dac931fb38d83493ab2652f – do you see anything breaking from these?

Notable changes:

- I changed the `module` in all `packages/*/package.json`  to `src/index.js`, which is needed otherwise during testing the dependencies get duplicate imports all over the place
- I disabled uglify b/c it broke with a parsing error on tko.js
- I removed `postinstall`  from `package.json` because it was causing an infinite yarn -> lerna -> yarn loop

[...]

Issue #6, opened 2016-09-18
Typescript support


brianmhunt:
From #1 , just so I can clean up relevant issues there.

----
IanYates commented on Jul 6
It really depends on whether we're talking about writing the library in TS or just making it friendly to consume from TS.

The latter is easy - we just need to make a type definition. If this library stays pretty close to KO in terms of a public API surface then it'll have a very similar definition file. I'd be happy to put something together as this takes shape.

If you're talking about writing the library in TS then that's a much bigger discussion. I personally really like it 😍 but I can understand that many don't.

----

brianmhunt commented on Jul 6
Consumption will be the goal since I don't know TS. :)

I wonder if there's a way to automate a type definition. What do they look like?
 @IanYates

----     
IanYates commented on Jul 6
Automate.. Not that I'm aware of.

Here are some definitions of things with which you're rather familiar

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/cc3d223a946f661eff871787edeb0fcb8f0db156/knockout-secure-binding/knockout-secure-binding.d.ts
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/cc3d223a946f661eff871787edeb0fcb8f0db156/knockout.amd.helpers/knockout-amd-helpers.d.ts
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/cc3d223a946f661eff871787edeb0fcb8f0db156/knockout.postbox/knockout-postbox.d.ts
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0b20db14dc8d546bd6e8e7b3bcaaf2289f63b55d/knockout.es5/knockout.es5.d.ts
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/cc3b657f2655c8c7cd290ff83dc052be637d8fe2/knockout/knockout.d.ts
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/17fe51d7edd2aab8917d6868b8b69e6014043559/knockout.punches/knockout.punches.d.ts

Also see http://www.typescriptlang.org/docs/handbook/knockout.html for an intro to TS using KO
and
http://www.typescriptlang.org/docs/handbook/writing-declaration-files.html for info about definition files straight from the horse's mouth
 👍 2

cervengoc:
Since TS 2, the best and most convenient way for handling typings is that each library provides its own `d.ts` files. This way there are no version conflicts, and the typings are coming together with the library itself in the npm package for example.

[...]

PR #39, opened 2017-10-15
Legacy browser support, better Monorepo usage

caseyWebb:
This is a first, although pretty comprehensive, pass at refactoring the monorepo that...

- uses lerna more idiomatically
- adds transpilation via tsc (TypeScript compiler) for legacy browser support (down to ES3)
- adds .es6, .min, and .es6.min builds
- (kind of) fixes the tests

### Idiomatic Lerna
The built browser bundle has been moved into the `packages` directory. Instead of using `rollup-plugin-includepaths`, packages are linked via `package.json => dependencies` (and `devDependencies` for test-only imports), which `lerna bootstrap` then does its magic on. `lerna bootstrap` has been registered as a postinstall hook at the root level, so installing deps and linking packages across the entire repo can be done with a `yarn install` in the root.

### Legacy Browser Support
Instead of Babel, the TypeScript compiler is used with the `allowJs` option and a target of `ES3`. It's faster, compiles down lower than babel (afaik, babel only supports ES5), and will allow for incremental adoption of TypeScript if desired.

### Multiple Builds
The browser bundle has 4 builds available: ES3, ES6, ES3+Minification, ES6+Minification.
Individual modules do not have minified builds, because they can only be consumed via JS bundlers and minification should be done last for best results (not to mention debugability).

### Modular Usage
TKO modules are passed through rollup, *but* other TKO modules are marked as external to prevent a bunch of duplication and are output in ES2015 module format and thus *only* supported with modern JS bundlers. Rolling them up renders faster builds for consumers wish to use individual modules, and allows transpilation without bloating the build process. Theoretically, we could create AMD/SystemJS bundles as well for script tag usage, but if you're going through the hassle of optimizing for weight, you're *probably* using ES2015 modules — at least you should be, IMHO of course.

The combined bundle on the other hand uses UMD format for universal consumption. AFAIK, tree-shaking won't take effect even if this is output in ES2015 module format without some refactoring, so this seems most reasonable for the moment.

Documentation for modular usage is very much needed. Currently the only way I could get it working was copying `packages/tko/src/index.js` to my own project and removing the parts I don't need.

### Tests
Jasmine was upgraded to an incompatible version at some point. This has been rolled back, so now the test failures are actual failures.


caseyWebb:
[x-ref #6]
[x-ref #21]

brianmhunt:
Awesome!  Looking forward to digging into this — will dig into it and comment this week.  Thanks @caseyWebb !

brianmhunt:
This is excellent.  There's a lot of much needed cleanup in here.

I was thinking of moving the tko `dist` to `packages/tko` too, so I'm glad you did that.

In general, `tko` is going to be a "reference implementation" of the modules, and not necessarily for export itself.  So copying it and modifying it for your needs is its essential purpose, though it'd be nice if there were an e.g. "tko.builder" package that brought all the pieces together in a more homogenous way.  The modularity documentation will go with the `tko` package.

The `knockout` package will stick to more conservative guidelines, though I haven't nailed down the exact delineation between Knockout and the tko reference implementation yet.

I think we can remove `~/.gitattributes`, since git-lfs is really more a pain than it's worth in this context.

The tests are written for Jasmine version 1.3, which is incompatible with Jasmine 2.0.  Jasmine 1.3 will be here for a long time; we use mocha+chai+sinon for some of the newer packages.

What's the difference between `npm run build` and `yarn build`?

The tests (`npm test`) aren't working for me; b/c of

> 19 10 2017 15:29:11.093:ERROR [preprocessor.rollup]: Error processing “spec/extenderBehaviors.js”

> Cannot read property 'slice' of undefined

> You need to include some adapter that implements __karma__.start method

Are you experiencing this too? Did you have thoughts on fixing them?

This is a pretty hefty commit but it's the way to go, so I'm ok merging this and fixing up anything it breaks, unless there's something that stands out as needing attention before that should happen.

Thanks @caseyWebb — This is a much needed and valuable patch!

caseyWebb:
A `tko.builder` would be superb. I'd be interested in helping design an API based on what is happening in `tko`.

+1 on git-lfs. I actually ran into an issue cloning the repo because I have an ssh alias for github which git-lfs couldn't resolve, and had to change the remote from `gh:knockout/tko` to the much more verbose `git@github.com:knockout/tko`. One of those things that's probably only an issue for a handful of people (or just me), but left me with a bad taste towards git-lfs in general.

As far as `npm run` vs `yarn`, it's exactly the same, but `yarn` allows omitting the `run` part of the command. Works with it, but works without it all the same. IMO, it's a bit less explicit, but more idiomatic. I only swapped it out in the docs to further emphasize that this project is based on `yarn`. I'm a fan of all-or-nothing usage for simplicity's sake.

Most of the tests pass for me, but there are still some errors. I'll have to double check to see what the errors are, but I don't believe it's those. To be honest, I'd really like to see a migration to jest instead of mocha if migrating the tests is a goal. It's extremely fast, and based on jasmine already.

There is definitely a lot going on in here. I began with trying to restrict changes to just the build, but revamping the lerna usage made that a *whole* lot easier and less overall code, so it seemed the right way to go.

AFAIK, there is nothing that needs to be done before merging this.

brianmhunt:
Awesome, thanks for the comment @caseyWebb 

Let's set up new issues and reference them:

1. start tko.builder
2. remove git-lfs
3. fix up testing
4. evaluate jest as an alt. test system

Just a quick merge & comment right now; I'm hoping to set aside some time for this soon, but feel free to get started with the issues.

PR #40, opened 2017-10-24
Remove git-lfs from project

caseyWebb:
Removes git-lfs, as mentioned in #39


brianmhunt:
I was wondering if we needed to remove any old lfs-distributed files from `dist/`, but I'm guessing not since they'll just appear as text files with the hash-pointer.  We can revise if it presents an issue, but this is great for moving forward, thanks @caseyWebb

Issue #50, opened 2017-12-23
Any other forward-ports


brianmhunt:
@mbest / @knockout/core / @knockout/extended  Does anyone see anything committed to Knockout 3.x line that's not been forward-ported?

1. after-render / childrenComplete
2. foreach update knockout/knockout#2324
3. any other plugins (e.g. your `switch/case` plugin)

mbest:
I thought I might see what's "missing" by running the latest Knockout tests against TKO, but there are a lot of failures. Any thoughts?

brianmhunt:
Thanks @mbest. That is an interesting methodology. I imagine there would be a lot of false negatives with that, as I had to manually modify about half the tests, but it still might give some insight. 

If you are set up for it, can you generate a list of the tests that fail? Then I can manually compare.

It might also be useful to compare the total list of tests.

mbest:
[Failing 236 specs.docx](https://github.com/knockout/tko/files/1585251/Failing.236.specs.docx)

mbest:
I "fixed" a bunch of tests by updating `useMockForTasks`. Here's what's left.

[Failing 162 specs.docx](https://github.com/knockout/tko/files/1585275/Failing.162.specs.docx)

brianmhunt:
Thanks @mbest 👍

mbest:
Merry Christmas!

[Failing 161 specs details.docx](https://github.com/knockout/tko/files/1586064/Failing.161.specs.details.docx)

Here's with the actual errors.

brianmhunt:
Awesome - very helpful! Thanks and merry Christmas 🎅

brianmhunt:
In #54 I've added the latest Knockout specs to the monorepo, as part of a `knockout` package in `packages/knockout`.

Gotten it down to 135 failing tests.

mbest:
That's cool!

brianmhunt:
Down to around 37 failing tests, most related to the foreach or the async children binding completion.

brianmhunt:
Subject to #68 and #69 #70 and comments and anything that fell through the cracks, I'm closing this as it appears complete.

PR #54, opened 2018-01-12
Add knockout proper, with spec

brianmhunt:
re. #50


brianmhunt:
Noting the outstanding issues in #68 and #69 #70

Issue #61, opened 2018-04-20
Up for grabs?


Sebazzz:
I love knockout and love to contribute to the next generation of Knockout. Is there room in the project to contribute?

brianmhunt:
Absolutely, @Sebazzz, thanks for reaching out .  Did you have anything in mind?

Sebazzz:
Good to hear. I'm not sure what readily tasks are available and what is required for them (for instance: are they thought out, or still need design, need additional reviewing etc).

brianmhunt:
@Sebazzz Thanks - the typescript support is one area where help would be valuable.

I've started a project https://github.com/knockout/tko/projects/2 to keep track of the high-level issues before a beta release.

Once we reach beta, we can start looking at all the additional features and API updates.  The plan would be to have another release not too long after we're satisfied with KO 4, with a number of the in-demand additional features & updates.

Being volunteer-only, it's going to take a while, but we'll get there. :)

brianmhunt:
If there's any areas where you're comfortable contributing, please don't hesitate to let us know.  I'm happy to have a call discussing it, too, if you think that'd help.

brianmhunt:
That's super-weird; thanks for trying/sorting that out, @Sebazzz .  Please share if there's any workarounds we can build in.

Sebazzz:
Deleting the symlink, ignoring node_modules appears to work. What is the purpose of that symlink?

brianmhunt:
I presume it was there so that there would be some node resolution that otherwise didn't work.

It may have already been resolved with yarn & lerna.

We can remove and if needed try another workaround.

brianmhunt:
Closing this, unless there's more to discuss/suggest here.

PR #62, opened 2018-04-25
Initial Typescriptification of tko.utils

Sebazzz:
**Work in progress**

As discussed in #6, initial Typescriptification of the core `tko.utils` package. This first pull request will set the direction for further Typescriptification, so let's get it right 😄 

Notes:
- Many of the lower-level DOM compatibility layers use a lot of `any` though I have attempted to minimalize it
- Added initial `tslint` config file for Typescript linting
- As discussed in #61, due to check-out issues on Windows, replaced the `node_modules` symlink
- Specs are not Typescriptified (imho not very important)

There are still some pending issues, which I am currently not being able to work out due to my unfamiliarity with rollup:
- Rollup always wants to take the `index.js` of a package, but in this case we have a `index.ts`. What can be done here? Possible solution: rename all `index.js` to `index.ts`.
- Currently no (typed) support for the `this` parameter in many of the array utils due to compilation errors. Typescript doesn't appear to correct method resolution.


mbest:
To make it easier to see the changes to each file, I think it would good to split this into two commits. The first just renames the files and the second has the actual changes.

Sebazzz:
You're right. I will do it evening.

caseyWebb:
Using `git mv` will preserve the file history, and not show the entire file as diffed (so 2 commits aren't needed).

Sebazzz:
It appears the rollup Typescript plugin is not really being maintained, and having multiple issues: https://github.com/rollup/rollup-plugin-typescript/issues/116

caseyWebb:
\[x-ref\]: https://github.com/Profiscience/knockout-contrib/issues/25#issuecomment-376030856

Just want to get clarification, are we going for TS or JS source for TKO? I'd started playing around with wiring up type-checking unobtrusively (only a handful of files), but I'm all for actually using TS the way it's intended to be used. If that is the case, I can start bringing in bits and pieces of knockout-contrib (see issue above).

brianmhunt:
This is great, thanks @Sebazzz .   There are a couple things I'd like to understand.

1. Style — there's a lot of munging of the changelog because of extra semicolons and such; I'd like to stick with StandardJS if possible.  Is there a TS equivalent? I googled but found nothing.

2. Typescript — I'd like to avoid Typescript as a hard dependency, and prefer to have the core written in ES6 with Typescript definitions "added on", so to speak.  The reason is that, basically, I don't trust Typescript to be around forever or to go the direction we need it to — I've only used it for a few hours and already found a couple serious bugs, one of which is a [WONTFIX](Microsoft/TypeScript#338).  I have trust issues 😄 , but I'm wary of the work needed to back out of TS if it becomes necessary —reminiscing on having had to spend a huge amount of time converting Coffeescript projects, I enjoy  the freedom of having ES6 as the lowest common denominator!  Is it possible to write the Typescript definitions in separate files, keeping the ES6 to itself?  In other words - does it have to be tightly coupled?

3. Core competencies — I/we may not be able to support it as well, at least not at the beginning.  TS isn't something I've used at all, and we don't want half the codebase in TS and the other half in ES6.  It's a pretty big conversion, and we'd need to know that we won't get 2/3rds of the way through only to lose the capacity.

None of these are blockers, but I feel it important to voice these concerns to make sure I understand the implications.  Typescript might be a great foundation for the TKO/KO future, but I'd at least like to know about the options before we commit to this fairly hefty change.

Thanks @Sebazzz — really appreciate your effort here, and hope to have these issues clarified soon!

caseyWebb:
1) Prettier. Let's use prettier.

2) It doesn't _have_ to be, but the type-checking will not be as good.

3) If `strict` is not enabled in `tsconfig.json`, you can change the file extension of a js file to ts, and it will work and compile file _as long as there aren't actual issues discovered_.

Sebazzz:
Don't forget that Typescript found already actual bugs in the codebase. I understand the doubt, but the benefits of type safety are already showing.

Sebazzz:
I prefer explicit semicolons over automatic semicolon insertion. The reason is that the latter may be [ambiguous](https://stackoverflow.com/q/34950322/646215) ([another example](http://lucumr.pocoo.org/2011/2/6/automatic-semicolon-insertion/)), so in some situations you are required to use semicolons. In my opinion it is better to be consistent and then use them everywhere. 

And this is personal taste: I'm able to parse code faster with semicolons because you can quickly scan where each statements ends. But perhaps this is old fashioned.

brianmhunt:
The main argument to not include semis at this point is because the diffs will have a poor signal-to-noise ratio i.e. any semantic differences will be indistinguishable from semicolon injection.

We can revisit the semicolon after, but for the JS→TS PRs we should keep the diffs crisp.

Sebazzz:
I understand that. You would still need to configure prettier to use the current coding standard as much as possible, otherwise the noise in the diff will be too high, it is good to be aware of that.

caseyWebb:
Prettier is less a linter, and more a formatter. Like StandardJS, it's opinionated and can't really be configured aside from semis or not. While ESLint/TSLint can do some formatting, it is the primary goal of prettier, so it's a lot more powerful. If you've ever worked in go, it's akin to gofmt.

I think it may be best to format the entire codebase in a separate PR to get all of the noise out of the way up front.

Sebazzz:
@caseyWebb Thanks. I'm unable to get actual compilation of Typescript working with the current configuration. So I could definitely use some help there :) It appears Rollup is currently configured to take an `index.js` file, while for `tko.utils` a `index.ts` file is now used.

brianmhunt:
Just a quick note - I've investigated a bit and am 100% on board with Typescript.

I'm pulling in the PR #50 to bring TKO up to speed with Knockout 3.5/6, and then this is on the list.

phillipc:
Probably obsolete because of #202. Some prototype-chains in @tko/observable need final attention.

Issue #66, opened 2018-06-14
Support JSS for styles


brianmhunt:
With JSX (#25) and [jss](https://github.com/cssinjs/jss) we should be able to fully compartmentalize the code for components.

I'm hoping to get this in for Beta-1, since once we're done catching TKO up to the mainline branch, this is relatively low hanging fruit.

mbest:
JSS looks cool. Related: knockout/knockout#1692

brianmhunt:
JSS supports TC39 (#22) observables, so I'd argue to properly get jss support we'd want TC39.  Experimenting now.

avickers:
If this is still a stumbling block, you could consider going with [Constructable Style Sheets](https://developers.google.com/web/updates/2019/02/constructable-stylesheets) instead of JSS.  They are live in Chrome, and Jason Miller made an attempt at [a tiny polyfill](https://codesandbox.io/s/j3qj7x8jmy), which appeared to work in FF and Safari when I tried it.

PR #73, opened 2018-06-15
Support JSX and Slots

brianmhunt:
Re #66


brianmhunt:
Cross-linking cssinjs/jss#728

brianmhunt:
Note that this was supposed to include `jss` support too, however the cssinjs/jss library was causing all sorts of build & karma problems.

Until a jss library is available in esm or otherwise, we'll put the jss part off.  In the mean time, it's simple to include in a custom component.

PR #76, opened 2018-06-24
22 tc39 observables

brianmhunt:
@mbest - You can see I've added in support for the TC39 `{next:...}` style of argument to `subscribe`. 

I use the argument-type (i.e. an object with a `next` method) to detect whether it's TC39-compliant call, in which case the current value of an observable is emitted immediately/synchronously.

It occurred to me that perhaps we want to have a different approach to this, such as adding a new event type (i.e. `state` instead of `change`).  Any thoughts?


mbest:
> It occurred to me that perhaps we want to have a different approach to this, such as adding a new event type (i.e. state instead of change). Any thoughts?

I suggest that it be done with a new event type, which could also work with regular subscriptions.

brianmhunt:
Thanks @mbest, I think that's a good idea.  Did you have any specifics in mind?

mbest:
`"state"` would work. Implementation-wise, it should emit the current value and then become a regular `"change"` subscription.

PR #92, opened 2019-03-15
TypeScript-ify

caseyWebb:
Includes #62, rebased w/ latest tko master, build process fixed.

_Note: I haven't actually reviewed the work by @Sebazzz, nor tried anything in the browser. This is just bringing that PR up to date and adding build support._

### TODO

- [ ] get tests passing
- [ ] finish refactoring into TS
- [ ] replace build process with something simpler


mbest:
Some comments:

* All functions should have a typed return value.
* Don't use `Function` as a type.
* For compatibility with Knockout 3.5.0, exported types should match https://github.com/knockout/knockout/blob/master/build/types/knockout.d.ts

brianmhunt:
Thanks @mbest — what's the reason for not using `Function` as a type?

Sebazzz:
It is not typed, you don't provide types for input parameters and output type. This is almost never what you want.

Met vriendelijke groet,
Sebastiaan Dammann
________________________________
Van: Brian M Hunt <notifications@github.com>
Verzonden: Friday, March 15, 2019 7:02:08 PM
Aan: knockout/tko
CC: Sebastiaan Dammann; Mention
Onderwerp: Re: [knockout/tko] TypeScript-ify (#92)


Thanks @mbest<https://eur01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fmbest&data=02%7C01%7C%7Ce4ada3b35aab458f138308d6a9705830%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C636882697305146832&sdata=1DDPLVEz2QrafU%2BSMrWYI7wEfLB0elA6EFIjZyINj7Y%3D&reserved=0> — what's the reason for not using Function as a type?

brianmhunt:
Thanks @Sebazzz — I'm not sure I understand.  Do you mean that `Function` is untyped, but the correct thing to do in TS is to have types for the function (e.g. `someFunction: (n: number) => any`)?

Sebazzz:
Yes, but you can model it as a type alias or interface for reusability and clarity if suitable.

For instance:

```
type SubscriptionCallback<T> = (newValue : T) => void;

// […]

interface Xxx<T> {
    void subscribe(callback: SubscriptionCallback<T>);
}
```

packages/utils/src/dom/disposal.ts
```diff
@@ -96,8 +96,7 @@ export function cleanNode (node: Node) {
 
     // ... then its descendants, where applicable
     if (cleanableNodeTypesWithDescendants[node.nodeType]) {
-      // cleanNodesInList((node as CleanableNodeWithDescendants).getElementsByTagName("*"))
-      cleanNodesInList((node as CleanableNodeWithDescendants).childNodes)
+      cleanNodesInList((node as any).getElementsByTagName('*'));
```
brianmhunt: Why not use childNodes?  How are these two options here different?

packages/utils/src/dom/virtualElements.ts
```diff
@@ -0,0 +1,249 @@
...
+// but it does give them a nonstandard alternative property called "text" that it can read reliably. Other browsers don't have that property.
+// So, use node.text where available, and node.nodeValue elsewhere
+import { emptyDomNode, setDomNodeChildren as setRegularDomNodeChildren } from './manipulation';
+import { removeNode } from './disposal';
+import { tagNameLower, isDomElement } from './info';
+import * as domData from './data';
+import options from '../options';
+
+// @ts-ignore
+import _let from '../../../tko.binding.core/src/let';
```
brianmhunt: I didn’t see where this was used. I’m sure it is, I just didn’t see where.

packages/utils/src/symbol.ts
```diff
@@ -0,0 +1,9 @@
+//
+// ES6 Symbols
+//
+
+export const useSymbols = typeof Symbol === 'function';
+
+export function createSymbolOrString(identifier: any) {
```
brianmhunt: I think identifier has to be a string

packages/utils/src/dom/disposal.ts
caseyWebb: @brianmhunt the latter is recursive (grandchildren nodes). I changed it to childNodes under the assumption it'd be the same, but it broke the unit tests.

packages/utils/src/symbol.ts
caseyWebb: Can we just use ES2015 Symbols directly and polyfill?

packages/utils/src/dom/disposal.ts
brianmhunt: Ah I see.  That’s unfortunate since childNodes is much faster .... but evidently incorrect 😀

packages/utils/src/symbol.ts
brianmhunt: Yes definitely.  That’s been on my todo forever

packages/utils/src/array.ts
```diff
@@ -0,0 +1,198 @@
+//
+// Array utilities
+//
+// Note that the array functions may be called with
+// Array-like things, such as NodeList.
+
+export const {isArray} = Array;
+
+// export function arrayForEach<T, TThis>(array: T[], action: (this: TThis, item: T, index: number, array: T[]) => void, thisArg: TThis): void;
+export function arrayForEach<T>(array: T[], action: (item: T, index: number, array: T[]) => void, thisArg?: any) {
```
Sebazzz: I see a lot of the same callbacks. It would be nice for DRY-ness to make a seperate type for this (even if only used internally):

    type ArrayUtilCallback<T, TResult=void> = (item: T, index: number, array: T[]) => TResult;

packages/utils/src/async.ts
```diff
@@ -0,0 +1,26 @@
+//
+// Asynchronous functionality
+// ---
+import { safeSetTimeout } from './error';
+
+// tslint:disable-next-line:ban-types
+export function throttle(callback: Function, timeout: number) {
```
Sebazzz: Need to check whether we can make this more specific than Function?

packages/utils/src/css.ts
```diff
@@ -0,0 +1,38 @@
...
+  let addOrRemoveFn: (className: string) => void;
+  if (!classNames) { return; }
+  if (typeof node.classList === 'object') {
+    addOrRemoveFn = node.classList[shouldHaveClass ? 'add' : 'remove'];
+    arrayForEach(classNames.match(cssClassNameRegex)!, (className: string) => {
+      addOrRemoveFn.call(node.classList, className);
+    });
+  } else if (typeof (node.className as any).baseVal === 'string') {
+        // SVG tag .classNames is an SVGAnimatedString instance
+    toggleObjectClassPropertyString(node.className, 'baseVal', classNames, shouldHaveClass);
```
Sebazzz: IE < 10 isn't a target anymore right? Perhaps redirect directly to `classlist` instead of attempting to feature-detect and workaround?

packages/utils/src/array.ts
brianmhunt: @Sebazzz Where should these sorts of types be defined?

Sebazzz: Since it is specialized to the utility functions in this file, I'd say in this file at the top.

packages/utils/src/async.ts
brianmhunt: It looks like `lodash` has a generic `Function` as well: https://github.com/lodash/lodash/blob/master/throttle.js (though I'm not sure the jsdoc is used for the type checking, I noted that the [throttle type at DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/lodash-es/throttle.d.ts) just re-exports)

packages/utils/src/css.ts
brianmhunt: Yeah, there's a ton of pruning we could do for non-old-IE, including this.

packages/utils/src/array.ts
brianmhunt: Got it.

Related question:  In the case of JSX, we convert accept a number of types in `anyToNode`, with appropriate delegation by a `switch` e.g. https://github.com/knockout/tko/blob/master/packages/utils.jsx/src/JsxObserver.js#L233

I was thinking of rewriting that switch in TS with something like the following (with most specific arguments to most general):

```js
anyToNode(value: Node) : Node { ... }
anyToNode(value: null|undefined|Error|Symbol) : Node { ... }
anyToNode(value: object) : Node { ... }
anyToNode(value: function) : Node { ... }
anyToNode(value: string) : Node { ... }
anyToNode(value: boolean|number|bigint) : Node { ... }
```

If I understand it correctly, Typescript essentially is going to compile this down to a `switch` anyway.  Is there a better way to write this, or is this a sound approach? (Even this fairly naive TS version looks _**much**_ better to me, certainly, than the JS function with a switch). 

Also, the `anyToNode` returns one of `HTMComment`, `HTMLElement`, or `Text` - what's the right return type? The more general `Node`?  Or the most specific in each case? Or a combination with e.g. `HTMLElement|HTMLComment|Text`?

brianmhunt:
I'm looking into converting `@tko/observable` and I'm not sure the best way to add types to an Observable, but there's a lot of questions about things that aren't in the handbook.  Here's what I've got for the JS:

```js
export function observable (initialValue) {
  function Observable () {
    if (arguments.length > 0) {
            // Write
            // Ignore writes if the value hasn't changed
      if (Observable.isDifferent(Observable[LATEST_VALUE], arguments[0])) {
        Observable.valueWillMutate()
        Observable[LATEST_VALUE] = arguments[0]
        Observable.valueHasMutated()
      }
      return this // Permits chained assignments
    } else {
            // Read
      dependencyDetection.registerDependency(Observable) // The caller only needs to be notified of changes if they did a "read" operation
      return Observable[LATEST_VALUE]
    }
  }

  overwriteLengthPropertyIfSupported(Observable, { value: undefined })

  Observable[LATEST_VALUE] = initialValue

  subscribable.fn.init(Observable)

    // Inherit from 'observable'
  Object.setPrototypeOf(Observable, observable.fn)

  if (options.deferUpdates) {
    deferUpdates(Observable)
  }

  return Observable
}
```

On the TS side:

```ts
type Observable = (...args: any[]) => void

export function observable (initialValue: any) : Observable {
  function Observable (...args: any[]) : Observable {
    if (args.length > 0) {
      if (Observable.isDifferent(Observable[LATEST_VALUE], arguments[0])) { // Property 'isDifferent' does not exist on type '(...args: any[]) => Observable'
        Observable.valueWillMutate()
        Observable[LATEST_VALUE] = arguments[0]
        Observable.valueHasMutated()
      }
      return this // this' implicitly has type 'any' because it does not have a type annotation.ts(2683)
    } else {
            // Read
      dependencyDetection.registerDependency(Observable) // The caller only needs to be notified of changes if they did a "read" operation
      return Observable[LATEST_VALUE]
    }
  }

  overwriteLengthPropertyIfSupported(Observable, { value: undefined })

  Observable[LATEST_VALUE] = initialValue

  subscribable.fn.init(Observable)

    // Inherit from 'observable'
  Object.setPrototypeOf(Observable, observable.fn)

  if (options.deferUpdates) {
    deferUpdates(Observable)
  }

  return Observable
}
```

These issues seem to mostly stem from the `observable` function returning an `Observable` function / instance.  It isn't apparent how Typescript is set up to deal with `function() { this }`, leaving me with a bunch of questions to bridge the gaping chasm that is my TS knowledge.  Any thoughts and assistance is most welcome and appreciated.

packages/utils/src/array.ts
Sebazzz: I'm not familiar enough with the compiler-side of things regarding JSX, perhaps @mbest knows?
As I understand it, Typescript does not [support overloading](http://www.typescriptlang.org/docs/handbook/functions.html#overloads) in the sense that compiler figures out the checks to make, and the handbook does not express it explicitly.

Regarding the last: Yes, a union is the way to go since you know what types might be returned.

Sebazzz:
Are you looking for [this](http://www.typescriptlang.org/docs/handbook/functions.html#this)? 

Any reason why `Observable` is not generic?

brianmhunt:
> Any reason why `Observable` is not generic?

Just because I don't yet know what generic means :)

Sebazzz:
> 
> 
> > Any reason why `Observable` is not generic?
> 
> Just because I don't yet know what generic means :)

Ah, right, excuse me 😉

So currently any value can be set into the `Observable`, but if it were an `Observable<T>` with accepting input and output values of type `T` instead of `any` you get a strongly typed observable: If you create a `Observable<string>` you can't put a `Person` in it accidently.

brianmhunt:
> > > Any reason why `Observable` is not generic?
> > 
> > 
> > Just because I don't yet know what generic means :)
> 
> Ah, right, excuse me 😉
> 
> So currently any value can be set into the `Observable`, but if it were an `Observable<T>` with accepting input and output values of type `T` instead of `any` you get a strongly typed observable: If you create a `Observable<string>` you can't put a `Person` in it accidently.

That's unbelievably satisfying. 🤣

packages/utils/src/array.ts
caseyWebb: Maybe a stupid question, why do we have these in the first place if TKO doesn't target IE8?

brianmhunt: @caseyWebb The `anyToNode` function is needed because it's taking user-based input and turning it into JSX; that input may be in the form of a JSX-like object (`{ elementName: string, attributes: object, children: array }`) or text, or something else that we nevertheless want to handle gracefully such as an observable or promise.

mbest:
See knockout/knockout#2458

packages/utils/src/array.ts
brianmhunt: @avickers Turns out that section of the docs is only about function type descriptions; Typescript doesn't actually do the overloading implementation.  https://github.com/Microsoft/TypeScript/issues/5658 . It should.  But it doesn't.

Sebazzz: That wouldn't be easy. Only for the most trivial cases overloading could be implemented.

skewty:
Is the TS code in this MR functional?  It's been over 3 months since a comment was made so I guess everyone got busy with life.

brianmhunt:
@skewty This isn't yet functional (as far as I can tell), but in the interim i've written around 50,000 lines of TS, so my skills are _almost_ up to where they probably need to be to deliver on this PR. 👍

Alongside documentation, converting to TS is my ambition for TKO this year.

PR #98, opened 2019-04-10
bind) Make ko.applyBindings work with nodeType 11 (Document Fragment)

avickers:
It should be possible to apply bindings to nodeType 11 (Document Fragment).

Among other cases, the shadowRoot of a Shadow DOM / Web Component is a document fragment.

Currently, it is necessary to create a wrapper div as a child of the shadowRoot and parent of the actual template.  I can see no reason why binding to the shadowRoot directly should throw an error.

### DIFF
--- a/packages/bind/src/applyBindings.js
+++ b/packages/bind/src/applyBindings.js
@@ -47,7 +47,7 @@ function getBindingProvider () {
 }
 
 function isProviderForNode (provider, node) {
-  const nodeTypes = provider.FOR_NODE_TYPES || [1, 3, 8]
+  const nodeTypes = provider.FOR_NODE_TYPES || [1, 3, 8, 11]
   return nodeTypes.includes(node.nodeType)
 }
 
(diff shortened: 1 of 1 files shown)


avickers:
The 3rd commit should probably be a separate pull request.  I'll split it off if you'd like, Brian.

@mbest, the nodeType checks are inherited from Knockout 3.  Can you think of a reason not to allow applying bindings to shadowRoot / Web Components?  Would you like a pull request?

It would allow for a clean way of solving this: https://github.com/knockout/knockout/issues/2426 
Helping KO to play better with 2015+ standards, as well as other libraries/frameworks that use custom elements/components.

mbest:
This makes sense. I'd expect tests, though.

brianmhunt:
In principle this makes sense, and I'm happy to plug in the changes for recognizing type-11 nodes, subject to testing that verifies the behaviour.

What's the purpose of the changes to `Identifier`?  I noted that they break CSP `eval-unsafe`, which is a hard requirement for some using TKO.

avickers:
The change to Identifier was only to shrink the package size.  The Regex is something like 10K characters, but CSP explains why it needs to be that way.  That commit is irrelevant to getting applyBindings to work with the shadowRoot.

brianmhunt:
@avickers Got it, thanks.  I wish there were something better than that Regex, but if there is I haven't seen it. 😞  Otherwise, this is just a verification test away from merging into the next alpha.

phillipc:
Adopted in #229

PR #111, opened 2020-04-22
Update build process

brianmhunt:


phillipc:
Is this topic already settled with #195?

packages/observable/src/subscribable.ts
```diff
@@ -0,0 +1,349 @@
...
+export const LATEST_VALUE = Symbol('Knockout latest value')
+
+type Predicate<T> = (v: T) => boolean
+
+
+type Tc39Callback<T> = {
+  next (value: T): void
+}
+
+export function subscribable<T> (this: KnockoutSubscribable<T>): KnockoutSubscribable<T> {
```
phillipc, line 27: Here are some nice ts ports, many we could take up it again after merging #202.

brianmhunt:
outdated now

Issue #130, opened 2021-04-15
Packages & Makefile update


brianmhunt:
There are a number of housekeeping issues that need some attention:

1. updates the package dependencies to the latest versions
2. look at how `leana` is used and whether it's still needed in light of support in `yarn` workspaces
3. look at whether we could use `esbuild` in place of `rollup`

We also need to bump the version number to something sane, which my be an "official" number e.g. `4.0.0`, even though that would give the impression that there's more "official" stability to the build than is actually warranted.

There are also a number of "bald" packages (e.g. those not in the `@tko` organization) that need to be marked by npm as deprecated e.g.

- https://github.com/knockout/tko.git

brianmhunt:
I'm moving towards esbuild and nom 7 because esbuild is faster and npm 7 includes workspaces and doesn't require extra utils (i.e. yarn).  I'm also removing the Makefile dependencies.

There's quite a bit of experimenting because the nom / node import/export system is rather challenging, from a package developer perspective.

Based on some research I think I've decided I'd like to export both ESM and CommonJS, so I'm compiling `dist/index.js` for cjs, I'm also going to put an `dist/index.mjs` in the root.


I had attempted to add `dist/index

brianmhunt:
## Complication: CommonJS + ESM

To keep things simple, at least initially, I'm going to only compile ESM.  Since TKO is in (very) low usage, we can aim for future users, and worry about CommonJS when there's demand.  In addition to working well with tree-shaking and hopefully modern versions of all the widely used bundlers, it should also work with Deno.

brianmhunt:
## Exports

I'm going to rely on the [`exports` field](https://nodejs.org/api/packages.html#packages_conditional_exports) of package.json to disambiguate the type to serve.  On the basis that it replicates `main` and `module` properties, I'm removing these (though they could be added back in if they prove useful).

brianmhunt:
## Package Priorities

### `index.js` as default ESM
I'm exporting ESM as  `src/* => dist/*` e.g. `dist/index.js` because esbuild adds `.js` extensions to the import e.g. `import { x } from './array'` => `'./array.js'`. 

### unbundled ESM
The dist is unbundled so that we don't get the double-import problem (e.g. a copy of `@tko/utils` compiled into every bundle that's imported).

### CommonJS
The Common JS is exported bundled as `dist/index.cjs`

### Result

Here's what it looks like:

```json
  "exports": {
    ".": {
      "require": "./dist/index.cjs",
[...]

PR #136, opened 2021-05-26
130 1 package update

brianmhunt:
Implementation of #130


brianmhunt:
This largely subsumes #111 but there's code in there we'll want to merge forward.

There's a good chance there's a ton of things that we'll want to additionally merge to fix this code-path up, but we've got to take a bit of a leap of faith and merge in this branch and start tackling things in smaller, more digestible PRs.

PR #140, opened 2021-06-23
Begin site setup for www tko io

brianmhunt:
This PR will create a couple sites that show how to use ESM + TKO, namely the minimal build and the tko.io website.

I was using very clever (read: bad) combinations of pug + Python, etc., when I really should've been eating our own dog food, so to speak.

I'm going to move over all the content from https://github.com/knockout/io.tko/ and use the styles with [JSS](https://cssinjs.org/?v=v10.6.0)†.

I also plan to add a Github action for deploying to e.g. Firebase Hosting (i.e. Fastly/CDN).

The objective of this demonstration will be to illustrate compartmentalization of the HTML (JSX) + CSS (JSS) and Javascript in one file, and inheritance.

Some Typescript T/JSX typings may come out of this, too.

† big shout-out to the JSS developers, it compliments TKO/JSX very well.


brianmhunt:
Note that we've content in:
- https://github.com/knockout/io.tko/tree/archive/2019/from-tko and
- https://github.com/knockout/io.tko/tree/archive/2016%2Foriginal

PR #142, opened 2021-07-03
examples) add the-time example

brianmhunt:
Experiment, per: https://codesandbox.io/s/tko-sample-upy4z


brianmhunt:
Superseded by the examples showcase (#291). Thanks for the original work here.

Issue #147, opened 2021-07-13
Add preventDefault to event handler options


danieldickison:
Because lambda (arrow functions) in bindings currently only support a single expression, the old pattern of returning `true` from a (e.g.) click handler is a bit inconvenient:
```
data-bind="click: function (event) {
    setSomeValue('foo');
    return true;
}"
```

Adding a `preventDefault` key to an event handler was discussed in #32 but it doesn't appear to have been implemented. Does this still sound like a good idea?

```
data-bind="click: {
    handler: (event) => setSomeValue('foo'),
    preventDefault: false
}"
```

brianmhunt:
@danieldickison I think this scratches an itch a few people will have. 😀

maskmaster:
Using a helper-function like
```html
data-bind="click: utils.preventDefault(event => setSomeValue('foo'))"
```
would be a more terse variant. The exact semantics from knockout could in this case be retained.

Of course, if `preventDefault` needs to be observable we might need to do it like
```html
data-bind="click: event => { setSomeValue('foo')); return $data.shouldPreventDefault; }"
```

A third alternative, and in my opinion more semantically correct, is to create your own binding handler "capturedClick" or something like that, where you describe the functionality instead of implementing/open-coding it in the view.

tscpp:
Isn't this what we have a viewModel for? Also, what about stopPropagation() and stopImmediatePropagation()?

brianmhunt:
I tend to think there are two "correct" solutions here.

#### 1. The best option for a system where you know what you need and have total control is what @maskmaster suggested:

> A third alternative, and in my opinion more semantically correct, is to create your own binding handler "capturedClick" or something like that, where you describe the functionality instead of implementing/open-coding it in the view.

No library changes would be needed for this.

#### 2. a preferable option for a "drop-in" pre-built library, when experimenting (e.g. before knowing _what_ works), or behavioural visibility is better in the binding, can be to have a parameterized system...as @tscpp noted:

> Also, what about stopPropagation() and stopImmediatePropagation()? These functions and preventDefault() is functions, not options.

in which case `stopPropagation` and `stopImmediatePropagation` ought to be parameters in the library as well.

There's nothing stopping folks from using a function e.g. `utils.preventDefault` but that doesn't require any library changes.

brianmhunt:
Closing this as I trust it's closed, but we can always re-open if there's more to discuss here.

PR #148, opened 2021-07-13
Add preventDefault option to event bindings

danieldickison:
Here's my attempt at #147 to add a `preventDefault` option to event bindings, as an alternative to having the handler function return a boolean.


packages/binding.core/spec/eventBehaviors.ts
```diff
@@ -62,12 +62,27 @@ describe('Binding: Event', function () {
...
   it('Should prevent default action', function () {
-    testNode.innerHTML = "<a href='http://www.example.com/' data-bind='event: { click: noop }'>hey</button>"
+    testNode.innerHTML = "<a href='http://www.example.com/' data-bind='event: { click: noop }'>hey</a>"
     applyBindings({ noop: function () {} }, testNode)
     triggerEvent(testNode.childNodes[0], 'click')
         // Assuming we haven't been redirected to http://www.example.com/, this spec has now passed
   })
 
+  it('Should allow default action by setting preventDefault:false', function () {
+    testNode.innerHTML = "<div data-bind='event: {click: test}'><a href='#' data-bind='event: {click: {preventDefault: obs}}'>hey</a></div>"
```
brianmhunt, line 72: Should probably test the static-case

packages/binding.core/src/event.ts
```diff
@@ -42,8 +46,9 @@ export const eventHandler = {
             handlerReturnValue = handler.apply(possiblyUpdatedViewModel, argsForHandler)
           }
         } finally {
-          if (handlerReturnValue !== true) {
-            // Normally we want to prevent default action. Developer can override this be explicitly returning true.
+          if (handlerReturnValue !== true && (preventDefault === undefined || unwrap(preventDefault))) {
```
brianmhunt, line 49: I find this logic hard to grok (it always has been).

Perhaps we could break out the `preventDefault === undefined || unwrap(preventDefault)` into a separate, descriptive variable.

I'm also ok with another variable name that _may_ be clearer e.g. `requireDefault` or `doNotPreventDefault` (and whose truthiness matches an `undefined` state, making it a bit easier to comprehend)

brianmhunt (changes requested):
Thanks for this, looks like there's a few minor things and this would be ok to merge.

packages/binding.core/src/event.ts
danieldickison: Another thought I had was to ignore the return value completely when preventDefault is specified in the handler. That might be clearer and also less surprising when a in inline lambda handler accidentally returns a Boolean value.

danieldickison: I made it so preventDefault in the descriptor takes precedence over the return value. Lmk what you think of that -- I think this will be the least surprising behavior while maintaining backwards compatibility.

packages/binding.core/src/event.ts
```diff
@@ -42,8 +46,12 @@ export const eventHandler = {
             handlerReturnValue = handler.apply(possiblyUpdatedViewModel, argsForHandler)
           }
         } finally {
-          if (handlerReturnValue !== true) {
-            // Normally we want to prevent default action. Developer can override this be explicitly returning true.
+          // preventDefault in the descriptor takes precedent over the handler return value
+          if (preventDefault !== undefined) {
+            if (unwrap(preventDefault)) { event.preventDefault() }
```
danieldickison, line 51: Are we still supporting IE <9? I skipped the `event.returnValue` since it shouldn't be necessary for any newer browsers.

nmocruz: Even IE11 shouldn't worth any extra work. This browser is a problem to everyone.

brianmhunt: Anything IE we can / should drop. TKO 4.1 won't guarantee anything IE, because it paralyzes growth. I think the baseline is going to be chrome 50+ (because of Unicode regular expressions), subject to others putting filling in the gaps (which will be possible on a modular basis I hope).

packages/binding.core/src/event.ts
brianmhunt: Yeah, I think the most expressive and intuitive is to have the explicit argument determine the behaviour, and override any implied return value. Ie. if someone says preventDefault is X who are we to argue? 😀

packages/binding.core/src/event.ts
danieldickison: Cool. I'll go ahead and remove the `returnValue` code from the existing path, too.

packages/binding.core/spec/eventBehaviors.ts
```diff
@@ -62,12 +62,37 @@ describe('Binding: Event', function () {
...
   it('Should prevent default action', function () {
-    testNode.innerHTML = "<a href='http://www.example.com/' data-bind='event: { click: noop }'>hey</button>"
+    testNode.innerHTML = "<a href='http://www.example.com/' data-bind='event: { click: noop }'>hey</a>"
     applyBindings({ noop: function () {} }, testNode)
     triggerEvent(testNode.childNodes[0], 'click')
         // Assuming we haven't been redirected to http://www.example.com/, this spec has now passed
   })
 
+  it('Should allow default action by setting preventDefault:false', function () {
+    testNode.innerHTML = "<div data-bind='event: {click: test}'><a href='#' data-bind='event: {click: {preventDefault: false}}'>hey</a></div>"
```
brianmhunt, line 72: We could also test `preventDefault: true` here, though the code path for that is identical to the `unwrap()` tested below so I'm not concerned.

brianmhunt (approved):
Looks good to me.

Sorry the CI tests don't run automatically on commit, but I trust the tests pass as expected.

Feel free to merge.

On a release plan, I'm working on a Github action for releases.  Note that it'll be with esbuild, which is not as well tested as the rollup version, so I'll be starting with 4.1.0... release tags.

danieldickison:
> Feel free to merge.

Am I able to merge this into the upstream myself? I assumed you'd need to initiate that merge. (Sorry, I'm not too familiar with the workflow in github.)

brianmhunt:
> Am I able to merge this into the upstream myself? I assumed you'd need to initiate that merge. (Sorry, I'm not too familiar with the workflow in github.)

I'm not sure. 😄 I merged it in any case, now.

PR #150, opened 2021-08-03
functionRewrite improvements

danieldickison:
* strip trailing semicolon from function body
* make `return` optional for side-effects-only functions
* log the args when rewriting
* include the binding text in error messages for any remaining deprecated `function` bindings so that these can be located more easily in the source

### DIFF
--- a/packages/utils.functionrewrite/spec/functionRewriteBehavior.ts
+++ b/packages/utils.functionrewrite/spec/functionRewriteBehavior.ts
@@ -14,7 +14,8 @@ describe('Function Rewrite Provider', function () {
       'x: function () { return {} }': 'x: () => { }',
       'x: function () { return "abc" }': 'x: () => "abc"',
       'stringLiteral: "hello", numberLiteral: 123, boolLiteralTrue: true, boolLiteralFalse: false, objectLiteral: {}, functionLiteral: function() { }, nullLiteral: null, undefinedLiteral: undefined': 'stringLiteral: "hello", numberLiteral: 123, boolLiteralTrue: true, boolLiteralFalse: false, objectLiteral: {}, functionLiteral: () => undefined, nullLiteral: null, undefinedLiteral: undefined',
-      'function (v) { return v.name + " (" + v.job + ")"; }': '(v) => v.name + " (" + v.job + ")"'
+      'function (v) { return v.name + " (" + v.job + ")"; }': '(v) => v.name + " (" + v.job + ")"',
+      'function () { foo(); }': '() => foo() && undefined'
     }
     const idempotents = [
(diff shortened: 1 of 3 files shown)


danieldickison:
I'm not sure how to fix the `Polynomial regular expression used on uncontrolled data` issue, but I don't think it really matters because we’re only parsing bindings written by the site developer, not those entered by end users.

packages/utils.functionrewrite/spec/functionRewriteBehavior.ts
```diff
@@ -14,7 +14,8 @@ describe('Function Rewrite Provider', function () {
       'x: function () { return {} }': 'x: () => { }',
       'x: function () { return "abc" }': 'x: () => "abc"',
       'stringLiteral: "hello", numberLiteral: 123, boolLiteralTrue: true, boolLiteralFalse: false, objectLiteral: {}, functionLiteral: function() { }, nullLiteral: null, undefinedLiteral: undefined': 'stringLiteral: "hello", numberLiteral: 123, boolLiteralTrue: true, boolLiteralFalse: false, objectLiteral: {}, functionLiteral: () => undefined, nullLiteral: null, undefinedLiteral: undefined',
-      'function (v) { return v.name + " (" + v.job + ")"; }': '(v) => v.name + " (" + v.job + ")";'
+      'function (v) { return v.name + " (" + v.job + ")"; }': '(v) => v.name + " (" + v.job + ")"',
+      'function () { foo(); }': '() => foo() && undefined'
```
brianmhunt, line 18: A safer way to return `undefined` is `void 0`, but no modern browser lets you overwrite `undefined` now so I think this is fine.

See e.g.: https://stackoverflow.com/questions/5716976

packages/utils.parser/src/Parser.ts
```diff
@@ -656,7 +656,7 @@ export default class Parser {
       case 'null': return null
       case 'undefined': return void 0
       case 'function':
-        throw new Error('Knockout: Anonymous functions are no longer supported, but `=>` lambdas are.')
+        throw new Error('Knockout: Anonymous functions are no longer supported, but `=>` lambdas are. In: ' + this.text)
```
brianmhunt, line 659: Smart

brianmhunt (approved):
@danieldickison Looks good! I just added a comment on the `undefined` vs `void 0` but it's innocuous nowadays.  I've approved, please feel free to merge (or let me know and I can do it if you prefer).

danieldickison:
@brianmhunt I figure `undefined` should be safe with the browsers we now support.

I don't have permissions on the repo so please merge it upstream. Thanks!

Issue #146, opened 2021-07-13
optionsText with a lambda value fails to resolve the function argument


danieldickison:
I tried a lambda (arrow-function) value for `optionsText`, following the [knockout docs’ example](http://knockoutjs.com/documentation/options-binding.html), which uses a `function` value:

```html
<select data-bind="
    options: possibleUsers,
    optionsText: (user) => formatUser(user),
    value: selectedUser
"></select>
```

This fails with:
> Message: The variable \&quot;user\&quot; was not found on $data, $context, or globals.

Specifying the `optionsText` as a function reference does work: `optionsText: formatUser`

Demo: https://jsfiddle.net/danieldickison/az2on6h0/4/

brianmhunt:
Good find, thanks. This would seem to be a problem with the `bindingContext` `lookup` function.

A PR with a failing test would be a good starting point.

Glad you've a workaround with the `formatUser`, in the interim.

danieldickison:
Tests in #151 and it looks like lambda parameter lists aren't being evaluated at all at the moment, and multiple parameters fail to parse. I've been looking at parser.ts but haven't quite wrapped my head around how all the parts fit together.

@brianmhunt do you think you would be able to look into this soon? If not, I'll continue developing a fix in the parser over the next couple of days.

Issue #65, opened 2018-05-24
Document Descendants and Children Complete


brianmhunt:
@mbest - I've been trying to implement the DescendantsComplete and ChildrenComplete events (i.e. pass all the unit tests), and it's been taking a bit longer than it should because TKO does a few internals differently now and I haven't fully figured out how these events should work.

It would help me a ton if you were able to put together a bit of documentation (either formally or informally here) that relayed how they act in the various circumstances.  It's clearly well thought out, but because the TKO internals are different the don't align exactly.

A high level description would go a long way, but some code-comments on the `AsyncCompleteContext` (why we need it, how it differs from a regular binding context) would go a long way and be much appreciated!

mbest:
`childrenComplete` is triggered as soon as bindings are completed on the node's children. *Completed* means that `applyBindings` was called and returned for each node. It's always triggered synchronously from the point of view of the code that applies bindings to the child nodes. It will be triggered again if the node's children are re-bound (usually after being cleared and re-rendered). It's not triggered if the node has no children.

`descendantsComplete` is triggered when all the node's descendant nodes are completed. If any of the descendant nodes indicates that it will complete its children asynchronously, the event will be delayed until that node's children are completed. In the case where all descendants are bound and completed synchronously, it will be triggered synchronously.

brianmhunt:
Thanks Michael.  Should `descendantsComplete` trigger repeatedly e.g. with `foreach` changes?

mbest:
> Should descendantsComplete trigger repeatedly e.g. with foreach changes?

I think it will since it's triggered by the `childrenComplete` event. I just noticed that I don't have any tests for these events and `foreach`. I'll have to work on that.

brianmhunt:
Got it, thanks.

Relatedly, in `tko` the `applyBindings` returns an instance of a class called [BindingResult](https://github.com/knockout/tko/blob/50-knockout/packages/tko.bind/src/BindingResult.js).  I'm mulling what it should return, but in essence I'm hoping it can expose the binding results, events, and errors in a way that helps debugging and makes KO easier to implement in certain use-cases.  Any thoughts on it?

mbest:
Here's more on how the nodes need to interact to trigger a `descendantsComplete` event:

1. A node that will complete asynchronously needs to mark itself as such; later it will need to clear that mark when it's complete.
2. A marked node is considered complete if it's removed or cleaned (the mark is cleared).
[...]

PR #151, opened 2021-08-03
Fix parsing and evaluation of lambdas with args and operator associativity

danieldickison:
Turns out parserBehavior.ts already had tests that were being skipped for lambdas with args. It references #65 but I'm not sure how that issue is related. This appears to be the root of issue #146 

This should also fix incorrect operator associativity (see  [comment 894513849](https://github.com/knockout/tko/pull/151#issuecomment-894513849) and #153).


danieldickison:
I've updated the PR with a fix — it parses an arg list in front of `=>` by taking the previously parsed Expression and turning it into a new Parameters node, which extends the context when the lambda is called. This involved adding a new `,` operator and dropping the `=>` operator precedence to the lowest possible.

It's a little bit hacky, but I think it works well given how the parser is organized. (In actual JS grammar, `=>` is not defined as an operator, so would require some reworking if we want to mirror that parsing.)

danieldickison:
In the course of testing this, I discovered another issue with the parser — it was incorrectly right-associating operators with equal precedence, and also mishandling (in a not totally clear way) for some deeply nested trees. I rewrote `Node.create_root` to use the shunting yard algorithm to handle precedence and updated tests as appropriate.

PR #155, opened 2021-09-02
Use == for comparing select option values for better ko compatibility

danieldickison:
Knockout uses `==` for comparing select option values against the value binding value, and this PR restores that behavior for better compatibility. This is critical if you have bindings like this:
```html
<select data-bind="value: typeID">
  <option value="1">foo</option>
  <option value="2">bar</option>
  ...
```

I'll try to make a test for this if I have time.

### DIFF
--- a/packages/utils/spec/utilsDomBehaviors.ts
+++ b/packages/utils/spec/utilsDomBehaviors.ts
@@ -123,3 +123,21 @@ describe('cloneNodes', function () {
     expect(childIsClone).toBe(true)
   })
 })
+
+describe('selectExtensions', () => {
+  beforeEach(jasmine.prepareTestNode)
+
+  it('should use loose equality for select value', () => {
+    const select = document.createElement('select')
+    select.innerHTML = `
+      <option value="42" selected>Forty-two</option>
+      <option value="84">Eighty-four</option>
+    `
+    testNode.appendChild(select)
+
+    expect(select.selectedIndex).toBe(0)
+    expect(utils.selectExtensions.readValue(select)).toBe('42')
+    utils.selectExtensions.writeValue(select, 84, true)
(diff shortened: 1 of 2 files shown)


mbest:
The problem with this change is that although `typeID` might start as a number, it will get updated to a string when the user selects something. Instead, set your options to have the numeric value: 

```
<select data-bind="value: typeID">
  <option data-bind="value: 1">foo</option>
  <option data-bind="value: 2">bar</option>
  ...
```

mbest:
I reopened this because the compatibility issue is still relevant.

brianmhunt:
Thanks @danieldickison -- what do you think of the more explicit equality test in my comment above?  I'd like to avoid linters complaining about the "evil twins" and I think the more explicit we can be the easier to maintain 🤞🏻

danieldickison:
Personally, I think just going with `==` along with a lint-suppression comment would be clearer than reimplementing that behavior explicitly, but that's a loosely-held opinion :)

Specifically, I think your solution should work fine, though maybe this bit:
```const blankEqual = optionValue === '' && value === undefined```
should be loosened up a bit to accept false/null/0 values:
```const blankEqual = optionValue === '' && !value```

brianmhunt:
😁  Hmm... I think the clarity of the code is important, but more important is defining and testing the conditions.

Do we want `0` and `false` to be considered "blank"?  I think `undefined` certainly, but `null` might also be a "positive" value.

@mbest any thoughts?

Not to drag the discussion out, but just want to be sure we end up with the expected behaviour.

nmocruz:
That why this 4 version is taking so much time. People takes 2 months to
decide something like this, and the decision is scale out to someone else.

mbest:
I prefer the more specific comparison.

PR #163, opened 2022-01-09
select-value-equality

brianmhunt:
- Use == for comparing select option values
- Add test for loose equality in selectExtensions


brianmhunt:
closes #155

PR #165, opened 2022-01-18
Make options.bindingGlobals empty by default

danieldickison:
This should make the default build safer by preventing access to things like document and fetch. For backwards compatibility, I think we need to keep `window` the bindingGlobals for the knockout build.


brianmhunt (approved):
:thumbsup:

PR #169, opened 2022-05-06
Re-enable knockout build tests and fix dummyTemplateEngine import

### LINKED ISSUE #162: Re-enable tests on builds/knockout
Disabled because the global `ko` is undefined and `dummyTemplateEngine` wasn't properly importing.

This may be a canary for other build issues since switching to esbuild, which we'll undoubtedly need to resolve sooner rather than later.

bcdanieldickison:
Fixes #162 though I'm not quite sure it's the right way.

### DIFF
--- a/builds/knockout/Makefile
+++ b/builds/knockout/Makefile
@@ -3,9 +3,6 @@ include ../../tools/build.mk
 
 iife-global-name := ko
 
-test:
-	@echo "Disabled pending fixes to build process"
-
 default::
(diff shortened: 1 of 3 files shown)


brianmhunt:
@mbest We were just trying to re-enable the testing of the compiled `knockout.js` but the cross-window testing went awry when we switched to `esbuild` (from `rollup`) — does this fix seem right to you?

mbest (approved):

PR #189, opened 2024-12-04
non sctrict operators compare strict

Auge19:
The non-strict equality-comparer == and != are comparing strict.

### DIFF
--- a/packages/utils.parser/src/operators.ts
+++ b/packages/utils.parser/src/operators.ts
@@ -42,8 +42,8 @@ const operators = {
   //    TODO: 'in': function (a, b) { return a in b; },
   //    TODO: 'instanceof': function (a, b) { return a instanceof b; },
   // equality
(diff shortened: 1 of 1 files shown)


brianmhunt:
IIRC This is by design, for backwards compatibility with Knockout.

I contemplated other operators e.g. `~==` and `~!=` that expose the non-strict versions.

brianmhunt:
I tend to agree. I'm not exactly sure where the strict equality came in, but one of the tenets of (T)KO is backwards compatibility, so I'll approve.

Before releasing a new version we'll need to ensure the tests work/pass.

brianmhunt:
It appears `make test` passes

image

I'll merge this now, and when I get a chance try to get a release out.  Will post when it's up.

brianmhunt:
@Auge19 this should be fixed in the _knockout_ build (not the _reference_ build), in the beta 1.6 release now.

PR #195, opened 2025-01-08
Dockercontainer and fixed deps for testing

Auge19:
closed accidentally in our repo, when merged into main

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@ NPX		:= npx
 NODE  	:= node
 NPM		:= npm
 LERNA	:= npx lerna
+DOCKER	:= docker
 
 # Some make settings
 SHELL := bash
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -22,12 +23,18 @@ all:: node_modules package-lock.json
...
+test-headless:
+	$(LERNA) exec --stream -- $(MAKE) test-headless
+
 ci:
 	$(LERNA) exec --stream --concurrency=1 -- $(MAKE) test-ci
 
 lint:
 	$(NPX) standard
 
+docker-build:
```
brianmhunt, line 35: I'd be inclined to add a `docker-test-headless` target too that rebuilds and runs the tests in headless docker image

brianmhunt (approved):
I'm ok with this, so I'll merge it in. 

My one concern is that the build & publish process is now different with this version of lerna, so I'll need some time/capacity to figure that out before releasing an updated version.

Thank you!

PR #196, opened 2025-01-09
Fixed typos in operators.ts and JsxObserver.ts

mcselle:
I found two typos while fixing Typscript compile errors in @Auge19 repo

### DIFF
--- a/packages/utils.parser/src/operators.ts
+++ b/packages/utils.parser/src/operators.ts
@@ -88,7 +88,7 @@ operators['--'].precedence = 16
 operators['&-'].precedence = 16
 
   // exponent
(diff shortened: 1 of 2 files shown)


packages/utils.parser/src/operators.ts
```diff
@@ -88,7 +88,7 @@ operators['--'].precedence = 16
 operators['&-'].precedence = 16
 
   // exponent
-operators['**'].precedent = 15
+operators['**'].precedence = 15
```
brianmhunt, line 91: would be good to have a test for this to prevent regression

brianmhunt (approved):
👍  thanks for this, much appreciated

PR #200, opened 2025-01-09
Fixed broken test after typo fix in JsxObservable.createInitialAdditions

mcselle:
after fixing that typo one of the test run into an error 
```
AssertionError: expected 'a<!--undefined-->b' to equal 'ab'`
```
because jsx[1] is undefined it results in a Comment Node with text undefined. 
so i changed the index

### DIFF
--- a/packages/utils.jsx/spec/jsxBehaviors.ts
+++ b/packages/utils.jsx/spec/jsxBehaviors.ts
@@ -370,7 +370,7 @@ describe('jsx', function () {
     const parent = document.createElement('div')
     const jsx = []
     jsx[0] = 'a'
-    jsx[2] = 'b'
+    jsx[1] = 'b'
     const jo = new JsxTestObserver(jsx, parent)
     assert.equal(parent.innerHTML, `ab`)
     jo.dispose()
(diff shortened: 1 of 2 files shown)


packages/utils.jsx/src/JsxObserver.ts
```diff
@@ -104,7 +104,7 @@ export class JsxObserver extends LifeCycle {
   }
 
   /**
-   * @param {HTMLElement|Comment|HTMLTemplateElement} parentNode
+   * @param {HMTLElement|Comment|HTMLTemplateElement} parentNode
```
phillipc, line 107: HTMLElement

brianmhunt:
@mcselle What typo / PR was it that broke the tests?

packages/utils.jsx/spec/jsxBehaviors.ts
```diff
@@ -370,7 +370,7 @@ describe('jsx', function () {
     const parent = document.createElement('div')
     const jsx = []
     jsx[0] = 'a'
-    jsx[2] = 'b'
+    jsx[1] = 'b'
```
phillipc, line 373: The test case is for sparse arrays, thats isn't more an array with a undefinied field in the middle.

phillipc (changes requested):

packages/utils.jsx/spec/jsxBehaviors.ts
phillipc: just change the assert to  `a<!--undefined-->b`

PR #202, opened 2025-01-12
Typescript) fix errors

phillipc:
As announced, we have started to make the TKO more type-safe. To get startpoint, we have set the tsconfig as softly as possible and continue to allow "any" types.

We (@Auge19 , @mcselle ) validate the typescript code with tsc. The current status is only 313 errors left. We began with 7410 errors. With this draft PR we would like to get feedback and start our own review. The goal is 0 errors with tsc. If this PR is ready and merged, Once this PR is done, we would like to continue with ESLINT.

All tests continue to pass.

PR #204, opened 2025-01-15
created github actions for building and compile checks

Auge19:
Two worklfows:
**Check TSC** runs after pull request is created. Runs `npx tsc` to check for any compile errors. Useful after `branch:strong_types` is merged.

**Build** runs after pushed to main branch. Executes a `make test` and stops if tests fail. Otherwise uploads content of `builds/knockout/dist` as artifacts. New release can be downloaded then in succeeded job as artifact. No manually build on local system is needed anymore.

### DIFF
--- a/.github/workflows/main-build.yml
+++ b/.github/workflows/main-build.yml
@@ -1,0 +1,42 @@
+# workflow for build and deploy
+name: Build
+
+on:
+  # Runs on pushes targeting the default branch
+  push:
+    branches: ["main"]
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [22.x]
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: 'npm'
+
+      - name: Install dependencies
+        run: npm install
+
+      - name: Run Build
+        run: make
+
+      - name: Run tests
+        run: make test-headless
+
+      - name: Upload build artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: tko
+          path: builds/knockout/dist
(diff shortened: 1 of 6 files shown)


Auge19:
package.lock must be added for action to work. Therefore `package-lock.json` must be removed from .gitignore to work. package.lock should not be ignored: [do-i-commit-the-package-lock-json](https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5)

phillipc:
I'm not sure about the "package-lock.json".. Your branch was also not synchronized with main and the ts-settings didn't fit . I also added another pr-github-action for test automation.

Checks fails as currently expected. See:
- #200 
- #202

phillipc:
I have set the tsc-pipeline to `continue-on-error`, so we can switch on it later in #202

phillipc:
#200 is merged, tests passed

.github/workflows/main-build.yml
```diff
@@ -0,0 +1,42 @@
...
+    branches: ["main"]
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
```
brianmhunt, line 16: It doesn't seem like a matrix is needed for these, unless there's a future use where it might make sense.  Since npm is just the runner, the version doesn't seem too important, 
so these workflows could be simplified by removing the matrix.

brianmhunt:
This looks great, thank you. Is it done/ready to merge? Need a review?  I had one minor comment, but I'd be ok merging this if it's looking like it's ready.

.github/workflows/main-build.yml
Auge19: you're right. This is for the future and does not have a use right now. I remember some problems running with node v23. With a matrix you could verify which node version is valid to use (but maybe fixed with the last commits, I don't check if the node-problem exists anymore).

I made a matrix for further test or check jobs, that may come

phillipc (approved):
Ready to merge

.github/workflows/main-build.yml
phillipc: later test setups without jquery could use the matrix configuration, too.

PR #206, opened 2025-01-28
value.ts: isInputElement is undefined

phillipc:
While working on #202 we found another bug:

`isInputElement` is an undefined property, `isInput` (L.56) will be correct. I have validate it with the original sources and the ko-issues [122](https://github.com/SteveSanderson/knockout/issues/122).

IE support isn't longer relevant these days. However, it just shouldn't be broken unconsciously.


brianmhunt (approved):
👍 looks good to me, please feel free to merge at will

phillipc:
Thanks @brianmhunt but i think i don't have the right to merge it: "Only those with [write access](https://docs.github.com/en/get-started/learning-about-github/access-permissions-on-github) to this repository can merge pull requests."

brianmhunt:
thanks @phillipc, noted

PR #207, opened 2025-02-15
fix) recovering jquery was broken

phillipc:
I found a always-false-condition and a stange self-assignment: `options.jQuery = options.jQuery`. 

After comparing it with Knockout.js, i think the solution is clear (see https://github.com/knockout/knockout/blob/45e86bf0347cf742ca359bab051ea82fdfae7a5d/src/binding/bindingAttributeSyntax.js#L567)

### DIFF
--- a/packages/bind/src/applyBindings.ts
+++ b/packages/bind/src/applyBindings.ts
@@ -379,8 +379,8 @@ export function applyBindingsToDescendants (viewModelOrBindingContext, rootNode)
 export function applyBindings (viewModelOrBindingContext, rootNode, extendContextCallback) {
   const asyncBindingsApplied = new Set()
   // If jQuery is loaded after Knockout, we won't initially have access to it. So save it here.
-  if (!options.jQuery === undefined && options.jQuery) {
-    options.jQuery = options.jQuery
+  if (!options.jQuery && globalThis.jQuery) {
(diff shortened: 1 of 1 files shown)


brianmhunt:
> These could be later steps or should i change this in this PR, too? In my opinion, this is more of a separate topic.

Thanks @phillipc it feels separate to me too

brianmhunt (approved):
👍 Looks good to me, thank you.

You _should_ be able to merge at will once approved but please let me know if not.

phillipc:
I would be happy to apply for writing privileges..

grafik

phillipc:
I have invested some time in the jquery topic, see https://github.com/Auge19/tko/tree/fix_jquery

After JQuery was reused, a lot came up:
 - triggerEvent doesn't work for 'change'-Events -> We should set useOnlyNativeEvents to 'true'
 - jQuery-Templates are not used if HTML5-<template>-tag is supported (all browser higher ie11)

After #202 we can tackle this topic. This PR will stay separate and can be merged.

brianmhunt:
Thanks @phillipc - I've invited you to the TKO Writers team, which should grant you permission to merge approved PRs, but if not please let me know. Thank you!

phillipc:
Thanks for the invitation, @brianmhunt

PR #208, opened 2025-02-28
Fix for hasFocus / checked binding if not applied via DataBindProvider

mcselle:
## Testcase

```js
[...]
const createElementWithHasFocusBinding  = {
  init: (element: HTMLElement, valueAccessor: () => any) => {
    let parent = $(element);
    let $hasFocus = valueAccessor();
    
    applyBindingsToNode(parent[0], { hasFocus: $hasFocus })
    }
  };
options.bindingProviderInstance.bindingHandlers.set("customBinding",  createElementWithHasFocusBinding)

testNode.innerHTML = `<input data-bind='customBinding: customProps' />`

var $hasFocus= observable(false)
applyBindings({customProps: $hasFocus}, testNode );
[...]
```
## Fix 

```js
//  in hasFocus.ts
[...]
var modelValue = valueAccessor(isFocused, {onlyIfChanged: true});
if (isWriteableObservable(modelValue) && (modelValue.peek() !== isFocused)) {
  modelValue(isFocused);
}
[...]
```

Found this bug while migrating from ko to tko, we have a custom LegacyBinding that applies bindings itself.
So its not a ``DataBindProvider`` which would otherwise trigger ``Parser.ConvertToAccessor()`` and therefore override the ``value`` function.

```js
// in parser.ts
  convertToAccessors (result, context, globals, node) {
    objectForEach(result, (name, value) => {
      if (value instanceof Identifier) {
        Object.defineProperty(result, name, {
          value: function (optionalValue, options) {
            const currentValue = value.get_value(undefined, context, globals, node)
            if (arguments.length === 0) { return currentValue }
            const unchanged = optionalValue === currentValue
            if (options && options.onlyIfChanged && unchanged) { return }
            return value.set_value(optionalValue, context, globals) // <------------------------ THIS is missing
          }
        [...]
```

I guess after #202 is merged it would be best to migrate the hasFocus / checked / etc. Bindings like the ``value`` Binding which already inherites from ``BindingHandler`` that basically contains the same logic as my quick fix

```js
// in BindingHandler.ts
get value () { return this.valueAccessor() }
set value (v) {
  const va = this.valueAccessor()
  if (isWriteableObservable(va)) {
    va(v)
  } else {
    this.valueAccessor(v)
  }
}
```

### DIFF
--- a/packages/binding.core/spec/hasfocusBehaviors.ts
+++ b/packages/binding.core/spec/hasfocusBehaviors.ts
@@ -3,8 +3,7 @@ import {
 } from '@tko/utils'
 
 import {
-    applyBindings
-    , applyBindingsToNode
+    applyBindings, applyBindingsToNode
 } from '@tko/bind'
 
 import {
@@ -22,8 +21,6 @@ import {
 import * as coreBindings from '../dist'
 
 import '@tko/utils/helpers/jasmine-13-helper'
-
-import $ from 'jquery';
 
 arrayForEach(['hasfocus', 'hasFocus'], binding => {
   describe(`Binding: ${binding}`, function () {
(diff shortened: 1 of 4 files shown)


phillipc (approved):
Good work! I know it was not easy to find.

phillipc:
@brianmhunt: I was involved in developing this solution. If this PR is okay for you, we can merge it.

packages/binding.core/spec/hasfocusBehaviors.ts
```diff
@@ -42,6 +44,33 @@ arrayForEach(['hasfocus', 'hasFocus'], binding => {
...
+          let $hasFocus = valueAccessor();
+
+          applyBindingsToNode(parent[0], { hasFocus: $hasFocus })
+        }
+      };
+      options.bindingProviderInstance.bindingHandlers.set("customBinding",  createElementWithHasFocusBinding)
+
+      testNode.innerHTML = `<input data-bind='customBinding: customProps' /><input />`
+
+      var $myVal = observable(false)
```
brianmhunt, line 60: prefer `const`.  The `$` prefix is generally reserved for "private" properties/accessors.  It's ok here, just noting.

phillipc:
@brianmhunt mcselle made all requested changes. Can I merge it now?

PR #225, opened 2025-12-23
TS) Enable "strictBindCallApply"

phillipc:
ts) set strictBindCallApply to true
ts) fixing result type of preprocessNode
api) required api change at TextMustacheProvider, see textInterpolationSpec

## Summary by CodeRabbit

* **Refactor**
  * Core provider/binding APIs now operate on generic nodes and consistently use explicit null to denote “no replacement”.
* **Bug Fixes**
  * Added runtime guards to skip non-Element nodes, preventing invalid element access and related errors.
* **Behavior**
  * Preprocessing now returns arrays for transformed nodes in more cases, standardizing results.
* **Tests**
  * Updated tests and helpers to expect null/array return semantics and adjusted templating/binding expectations.
* **Chores**
  * Enabled stricter TypeScript checks and tightened several internal type annotations.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/packages/binding.template/spec/foreachBehaviors.ts
+++ b/packages/binding.template/spec/foreachBehaviors.ts
@@ -944,7 +944,9 @@ describe('Binding: Foreach', function () {
   it('Can modify the set of top-level nodes in a foreach loop', function () {
     options.bindingProviderInstance.preprocessNode = function (node) {
       // Replace <data /> with <span data-bind="text: $data"></span>
-      if (!(node instanceof Element)) return null
+      if (!(node instanceof Element)) {
+        return null
+      }
 
       if (node.tagName && node.tagName.toLowerCase() === 'data') {
         const newNode = document.createElement('span')
--- a/packages/binding.template/spec/templatingBehaviors.ts
+++ b/packages/binding.template/spec/templatingBehaviors.ts
@@ -540,8 +540,9 @@ describe('Templating', function () {
       }
 
       override nodeHasBindings(node: Node, bindingContext?: BindingContext) {
-        if (node instanceof Element)
+        if (node instanceof Element) {
           return node.tagName == 'EM' || originalBindingProvider.nodeHasBindings(node, bindingContext)
+        }
         return false
       }
       override getBindingAccessors(node: Node, bindingContext?: BindingContext) {


packages/bind/src/arrayToDomNodeChildren.ts
```diff
@@ -205,7 +205,7 @@ export function setDomNodeChildrenFromArrayMapping<T = any>(
                   itemsForBeforeRemoveCallbacks.push(mapData!)
                 }
               }
-              if (mapData) {
+              if (mapData && mapData.mappedNodes) {
```
brianmhunt: can shorten to `mapData?.mappedNodes`

packages/binding.template/spec/foreachBehaviors.ts
```diff
@@ -944,6 +944,8 @@ describe('Binding: Foreach', function () {
   it('Can modify the set of top-level nodes in a foreach loop', function () {
     options.bindingProviderInstance.preprocessNode = function (node) {
       // Replace <data /> with <span data-bind="text: $data"></span>
+      if (!(node instanceof Element)) return null
```
brianmhunt: I prefer to keep return statements bound in braces i.e. `{ return null }` > `return null` to avoid the accidental newline breaking the condition, dangling else.  Not a big issue, just noting my style preference - particularly not an issue here in a `spec/` file :)

packages/binding.template/spec/nativeTemplateEngineBehaviors.ts
```diff
@@ -141,17 +141,25 @@ describe('Native template engine', function () {
...
+        items: ObservableArray<any>
+
+        constructor() {
+          this.items = observableArray([
+            { children: observableArray(['A1', 'A2', 'A3']) },
+            { children: observableArray(['B1', 'B2']) }
+          ])
+        }
+
+        invocationCount = function () {
```
brianmhunt: the `this` inside `function` will be the `this` of the anonymous `function ()`, solved with a method or lambda e.g.

```ts
invocationCount () { ++this.invocations } 
// or
invocationCount = () => ++this.invocations
```

I would expect that invocations will be `0` no matter what here.  It's not clear from the code review whether invocations being > 0 is validated.

packages/provider.virtual/src/VirtualProvider.ts
```diff
@@ -13,8 +13,8 @@ export default class VirtualProvider extends BindingStringProvider {
    * Convert <ko binding='...'> into <!-- ko binding: ... -->
    * @param {HTMLElement} node
    */
-  override preprocessNode(node: Element): Node[] | undefined {
-    if (node.tagName === 'KO') {
+  override preprocessNode(node: Node): Node[] | null {
+    if (node instanceof Element && node.tagName === 'KO') {
```
brianmhunt, line 17: this could be a duck-type e.g. `if ('tagName' in node && node.tagName === 'KO')`, which is more generic - but practically speaking probably doesn't matter here.

packages/utils/src/bind-shim.ts
```diff
@@ -1,6 +1,7 @@
 if (!Function.prototype['bind']) {
   // Shim/polyfill JavaScript Function.bind.
   // This implementation is based on the one in prototype.js
+  // Old browsers (IE7 and earlier) do not implement Function.bind.
```
brianmhunt, line 4: practically speaking we've got bigger issues for folks on IE7 lol

packages/binding.template/spec/nativeTemplateEngineBehaviors.ts
brianmhunt: @coderabbit please review this comment ^^ and comment on the context and the impact

phillipc:
Coderabbit drags me down the rabbit hole. The PR is getting bigger. I have also fixed the provider typings.

packages/utils/src/bind-shim.ts
phillipc: I'm looking forward to the moment when we can delete all of this. Currently, I'm tagging things like this so I can search for them using IE.

packages/provider.virtual/src/VirtualProvider.ts
phillipc: "Instanceof" has the advantage that the TypeScript type tracking is corrected. From here on, ts understands that it is an element.

packages/bind/src/arrayToDomNodeChildren.ts
phillipc: Done

packages/provider.native/src/NativeProvider.ts
```diff
@@ -31,8 +31,8 @@ export default class NativeProvider extends Provider {
    * There can be only one preprocessor; when there are native bindings,
    * prevent re-entrance (and likely XSS) from the `{{ }}` provider.
    */
-  override preprocessNode(node: Element) {
-    return node[NATIVE_BINDINGS] ? node : null
+  override preprocessNode(node: Node): Node[] | null {
+    return node[NATIVE_BINDINGS] ? [node] : null
   }
```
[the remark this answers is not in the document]
phillipc: This section in applybindings.ts was not been modified. The return value was ignored here before as well.

phillipc:
> > packages/provider.multi/src/MultiProvider.ts (2)> `51-58`: **Add explicit return type and return `null` instead of `undefined`.**
> > > According to the PR objectives, `preprocessNode` should return `Node[] | null`, but the implementation implicitly returns `undefined` when no provider returns `newNodes`. Additionally, the method lacks an explicit return type annotation.
> > > 
> > > 🔎 Proposed fix
> > > ```diff
> > > -override preprocessNode(node: Node) {
> > > +override preprocessNode(node: Node): Node[] | null {
> > >    for (const provider of this.providersFor(node)) {
> > >      const newNodes = provider.preprocessNode(node)
> > >      if (newNodes) {
> > >        return newNodes
> > >      }
> > >    }
> > > +  return null
> > >  }
> > > ```
@coderabbitai: These are relevant topics in the current TKO codebase. Please find more and create a new PR to correct them. 🫡

packages/binding.template/spec/nativeTemplateEngineBehaviors.ts
phillipc: fixed -> `invocationCount() {
          return ++this.invocations
        }`

phillipc:
@brianmhunt: I'm finished with the pull request. If you follow the existing code, then `preprocessNode(node: Node): Node[] | null` was always correct. Please provide a final review. I'll be happy if the pull request becomes smaller.

packages/provider/src/Provider.ts
```diff
@@ -14,6 +14,8 @@ export interface ProviderParamsInput {
   providers?: any[]
 }
 
+export type BindingAccessors = { [name: string]: Function }
```
brianmhunt, line 17: - `Function` is a broad definition, would be nice to narrow this type if we could.  Not a blocker, just noting.

packages/binding.template/spec/templatingBehaviors.ts
```diff
@@ -539,10 +539,14 @@ describe('Templating', function () {
         return [1, 3, 8]
       }
 
-      override nodeHasBindings(node: Element, bindingContext?: BindingContext) {
-        return node.tagName == 'EM' || originalBindingProvider.nodeHasBindings(node, bindingContext)
+      override nodeHasBindings(node: Node, bindingContext?: BindingContext) {
+        if (node instanceof Element)
+          return node.tagName == 'EM' || originalBindingProvider.nodeHasBindings(node, bindingContext)
```
brianmhunt, line 544: if (condition) followed by newline is dangerous.  Better to always use braces.  Not a blocker, but noting.

packages/binding.template/spec/foreachBehaviors.ts
phillipc: done

packages/binding.template/spec/templatingBehaviors.ts
phillipc: Done

PR #234, opened 2026-01-13
Add knip linter and fix all unused dependencies and exports

phillipc:
Knip finds and fixes unused dependencies, exports and files in your JavaScript and TypeScript projects.

In this PR we configured Knip and fixed all findings.

## Summary by CodeRabbit

* **Chores**
  * Added CI checks for package dependency health and new make targets; introduced dependency-analysis tooling and config; removed many redundant helper dependencies across packages.

* **Refactor**
  * Tightened TypeScript typings and reduced public API surface; converted several exports from default/public to internal or named-only.

* **Build**
  * Updated TypeScript compiler settings and ESLint config to preserve module syntax and widen ignored files.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,6 @@
         "esbuild": "^0.28.0",
         "eslint": "^10.2.0",
         "fs-extra": "^11.3.2",
-        "globals": "^17.4.0",
         "jquery": "^4.0.0",
         "karma": "^6.4.4",
         "karma-chrome-launcher": "^3.2.0",
@@ -9670,19 +9669,6 @@
       },
       "engines": {
         "node": ">=10.0"
-      }
-    },
-    "node_modules/globals": {
-      "version": "17.4.0",
-      "resolved": "https://registry.npmjs.org/globals/-/globals-17.4.0.tgz",
-      "integrity": "sha512-hjrNztw/VajQwOLsMNT1cbJiH2muO3OROCHnbehc8eY5JyD2gqz4AcMHPqgaOR59DjgUjYAYLeH699g/eWi2jw==",
-      "dev": true,
-      "license": "MIT",
-      "engines": {
-        "node": ">=18"
-      },
-      "funding": {
-        "url": "https://github.com/sponsors/sindresorhus"
       }
     },
     "node_modules/globalthis": {


packages/utils.jsx/src/JsxObserver.ts
```diff
@@ -464,4 +464,4 @@ export class JsxObserver extends LifeCycle {
   }
 }
 
-export default JsxObserver
+//Duplicated export default JsxObserver
```
brianmhunt: just remove this line, no need to comment it out

packages/provider/src/Provider.ts
```diff
@@ -55,7 +55,8 @@ export default class Provider {
   preprocessNode(node: Node): Node[] | null {
     return [node]
   }
-  postProcess(/* node */) {}
+
+  //TODO unused hook? postProcess(/* node */) {}
```
brianmhunt, line 59: I think leave `postProcess` there, it is meant to be overloaded, and the no-op serves for typing Provider generally.  Unless it's never used, in which case we can probably take it out (but I _feel_ like it was used, so we should be sure)

packages/observable/src/observableArray.ts
```diff
@@ -12,7 +12,7 @@ import { trackArrayChanges } from './observableArray.changeTracking'
 
 import type { Observable, ObservableFunctions } from './observable'
 
-export type MaybeObservableArray<T = any> = T[] | ObservableArray<T>
+//TODO Currently unused --> export type MaybeObservableArray<T = any> = T[] | ObservableArray<T>
```
brianmhunt, line 15: this is definitely used as an ambient declaration like KnockoutObservable, but perhaps not this version here.

brianmhunt, line 15: this is definitely used as an ambient declaration like KnockoutObservable, but perhaps not this version here.

Which is to say, I don't know if we can delete it, but lets not since it seems way out of scope for Knip

packages/computed/src/proxy.ts
```diff
@@ -88,12 +88,15 @@ export function proxy(object) {
   return proxy
 }
 
+/** @internal */
```
brianmhunt, line 91: I don't know what this is or why we'd add it, so lets skip it unless there's a clear need.

If something is internal it should be `private` or `#private`.  Also seems out of scope for Knip

packages/computed/src/computed.ts
```diff
@@ -67,7 +67,7 @@ export type ComputedReadFunction<T = any, TTarget = void> =
   | Computed<T>
   | ((this: TTarget) => T)
 export type ComputedWriteFunction<T = any, TTarget = void> = (this: TTarget, val: T) => void
-export type MaybeComputed<T = any> = T | Computed<T>
+// TODO: Currently unused -> export type MaybeComputed<T = any> = T | Computed<T>
```
brianmhunt, line 70: also definitely used.

.github/workflows/dependency-cycle.yml
```diff
@@ -0,0 +1,33 @@
...
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: 24
+          cache: 'npm'
+
+      - name: Install dependencies
+        run: npm install
```
brianmhunt: this should be a `make` target because we almost certainly want to switch to bun and not have hard-coded npm references all over the leafs.

packages/utils/src/interfaces.ts
```diff
@@ -0,0 +1,107 @@
...
+
+export type BindingAccessors = { [name: string]: Function }
+
+/**
+ * Public interface for minimal describing a binding provider.
+ *
+ * Providers are responsible for detecting bindings on DOM nodes and
+ * producing binding-accessor objects that the binding system consumes.
+ */
+export interface IProvider {
```
brianmhunt: the authority and single source of truth is Provider.    We dont want a redundant arbitrary interface that duplicates the class definition that we already have. No IAnything.

packages/utils/src/interfaces.ts
```diff
@@ -0,0 +1,107 @@
...
+import { extend, objectForEach, objectMap } from './object'
+import type { Options } from './options'
+import { createSymbolOrString } from './symbol'
+
+export interface IBindingHandlerObject {
+  set(nameOrObject: string | object, value?: string | object): void
+  get(nameOrDotted: string): any
+}
+
+export interface IArrayAndObjectUtils {
```
brianmhunt: this is a hard-coded subset of:

```ts
import * as array from 'array'
typeof array
```

but its not clear what benefit it adds, or what it has to do with Knip.

packages/utils/src/options.ts
```diff
@@ -17,7 +16,7 @@ export class Options {
   bindingStringPreparsers: BindingStringPreparsersFunction[] = []
 
   // Reference to the own knockout instance
-  knockoutInstance: KnockoutInstance | null = null
+  knockoutInstance: IKnockoutUtilsInstance | null = null
```
brianmhunt: I feel like this is both misleading and incorrect and will break a lot.  Let's take out all the I* things, it seems unrelated/unnecessary.

packages/provider/src/BindingHandlerObject.ts
```diff
@@ -1,6 +1,7 @@
 import { options } from '@tko/utils'
+import type { IBindingHandlerObject } from '@tko/utils'
 
-export default class BindingHandlerObject {
+export default class BindingHandlerObject implements IBindingHandlerObject {
```
brianmhunt: we're never going to have anything else that implements this, and now we have two sources of truth, making this fragile, redundant, and confusing.

packages/builder/src/Builder.ts
```diff
@@ -170,15 +160,15 @@ const utils: Utils = {
   unwrapObservable: unwrap
 }
 
-export type KnockoutInstance = {
+export type KnockoutInstance = IKnockoutUtilsInstance & {
```
brianmhunt: I'm unclear why we would do this.  Also we don't want a novel naming convention I* that's inconsistent with everything else in the project.  We can just call it `KnockoutUtils`.  It can also be a `class`.

packages/utils/src/interfaces.ts
phillipc: I'm with you. I proceeded this way to break the cyclical packages dependency. Utils <-> Provider and Bind. Any other ideas?

packages/computed/src/proxy.ts
phillipc: The reason for this is the export. In my opinion, the export only happens for testability in the same package. @internal is typescript sugar for such cases. https://www.typescriptlang.org/tsconfig/#stripInternal

Knip recognize this, too. And understand in this way that this is not an unnecessary export.

packages/observable/src/observableArray.ts
phillipc: Good to discuss. That's why I only commented on it. I can change it to @lintignore

Currently, Knip detected that this type def is unused

packages/provider/src/Provider.ts
phillipc: Okay, I will remove it. This hook is never called.

packages/utils/src/interfaces.ts
phillipc: It also breaks the cyclic reference between Builder.ts and Options.ts...

packages/provider/src/Provider.ts
phillipc: No call, no override, only definited as noop:
grafik

packages/observable/src/observableArray.ts
phillipc: We have ported this type from the official knockout.d.ts to tko, but never used it.

packages/computed/src/computed.ts
phillipc: Tell me where.. it is ported from the official ko-types, but unused in tko. Fun fact, it is also unused in the official knockout.d.ts, see https://github.com/knockout/knockout/blob/45e86bf0347cf742ca359bab051ea82fdfae7a5d/build/types/knockout.d.ts#L224

grafik

.github/workflows/dependency-cycle.yml
phillipc: done, duplicated step

packages/builder/src/Builder.ts
phillipc: I* is removed. We need another solution for the IProvider.

packages/provider/src/BindingHandlerObject.ts
phillipc: I've removed IBindingHandlerObject for now and replaced it with "any" in IProvider. This way, we're only dealing with the minimal problem with IProvider. Any idea?

Here again is the initial situation:

grafik

If this is corrected – then type cycle. In this case, everything still works because there are type imports. However, that's not correct.

packages/provider.mustache/src/AttributeMustacheProvider.ts
```diff
@@ -78,9 +78,9 @@ export default class AttributeMustacheProvider extends Provider {
     }
   }
 
-  getPossibleDirectBinding(attrName: string | number) {
+  getPossibleDirectBinding(attrName: string) {
     const bindingName = this.ATTRIBUTES_BINDING_MAP[attrName]
-    return bindingName && this.bindingHandlers.get(attrName)
+    return bindingName && this.bindingHandlers.get(attrName) //FIXME this.bindingHandlers.get(bindingName) ?
   }
```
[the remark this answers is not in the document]
phillipc: @coderabbitai Please create an issue for that. We need to fix it separately and also require a test case.

knip.json
```diff
@@ -0,0 +1,20 @@
...
+  "$schema": "https://unpkg.com/knip@5/schema.json",
+  "ignoreExportsUsedInFile": {
+    "interface": true,
+    "type": true
+  },
+  "tags": [
+    "-lintignore",
+    "-internal"
+  ],
+  "ignoreDependencies": ["@types/mocha", "chai", "electron", "esbuild", "fs-extra", "karma", "karma-[.]*", "mocha", "nyc", "prettier", "sinon", "lerna"],
```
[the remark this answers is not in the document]
phillipc: Done

packages/utils.jsx/src/JsxObserver.ts
phillipc: Done

packages/utils/src/interfaces.ts
phillipc: We need a type structure like this part for defining the return type of builder.create in builder.ts. This is a subset of some utils classes.

packages/utils/src/options.ts
phillipc: Done, changed to `KnockoutUtils`

packages/utils/src/interfaces.ts
phillipc: Probably an abstract base class? ProviderBase? (I wouldn't want to change the package structure.)

phillipc:
@brianmhunt: Summery, Knip detected some unused exports and missing dependencies at package.json level between the mono-repo-projects. While I adds the missing project references, some cyclical references were uncovered. 

To resolve that I have add some interfaces in utils. Now, I have refactor this PR to one last "IProvider" and "KnockoutUtils". I* notations are not typical typescript naming style, we can change it perhaps to an abstract base class like ProviderBase, but this is only an easy stylistic aspect.

You are welcome with a better idea to solve this problem. To put this in perspective, all of this is about the path to well-defined TKO types.

phillipc:
@brianmhunt please re-review this pr
I have removed all I*-notations. Thanks to knip, we now have a cycle-free, well-defined mono repo.

packages/computed/src/computed.ts
M-Kirchhoff: We should probably just remove it then

eslint.config.js
```diff
@@ -1,30 +1,29 @@
...
+const tseslint = require('typescript-eslint')
 
 
 /** @type {import('eslint').Linter.Config[]} */
-export default [
+module.exports = [
   pluginJs.configs.recommended,
   ...tseslint.configs.recommended,
-  { ignores: ['**/*.{js,mjs,cjs}', 'builds/**/*'] },
+  { ignores: ['**/*.{js,mjs,cjs}', 'builds/**/*', '**/*.d.ts'] },
```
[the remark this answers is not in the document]
phillipc: TKO is written in TypeScript. The last JS files are static-legacy or build scripts.

knip.json
```diff
@@ -0,0 +1,21 @@
...
+    "-internal"
+  ],
+  "ignoreBinaries": ["make"],
+  "ignoreDependencies": ["@types/mocha", "chai", "electron", "esbuild", "fs-extra", "karma[.]*", "mocha", "nyc", "prettier", "sinon", "lerna"],
+  "workspaces": {
+    "packages/*": {
+      "entry": "src/index.ts",
+      "project": "src/*.ts",
+      "ignore": ["/helpers/**"]
+    }
```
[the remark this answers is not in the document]
phillipc: fixed

PR #292, opened 2026-04-09
fix #290 - overloadOperator

### LINKED ISSUE #290: Error parsing Binding expressions with Operators == and != in reference Build
### Description:

Using bindings with operators `==` or `!=` (e.g. `<div data-bind="if: value() != null">`) will throw an exception “unexpected nodes remain in shunting yard output stack” while trying to parse.

After inspecting the code, it looks like the == and != operators are overridden in the reference build, but their precedence is not preserved/assigned on the overridden operator functions. This causes the shunting-yard reduction to leave pending operators, which then triggers the parser error.

### Expected behavior:
Expressions with `==` or `!=` should parse normally.

### Actual behavior:
Bindings using `==` or `!=` throw a parser error at runtime.

### Environment:

TKO reference build: `4.0.0` (npm package @tko/build.reference)

phillipc:
fix: add precedence parameter to overloadOperator and update usage in index.ts #290

## Summary by CodeRabbit

* **Bug Fixes**
  * #290 

* **New Features**
  * Added a new configuration option to toggle strict equality behavior for loose-equality operators (defaults to off).

* **Refactor**
  * Removed the runtime API for registering operator overloads; loose-equality behavior is now controlled by the new option rather than dynamic registration.


brianmhunt:
Could you split this into two PRs? The typo fix in `value.ts` (`propertyChangeFired` → `propertyChangedFired`) and the operator overload refactor (`overloadOperator` → `options.overloadEvilTwins`) are unrelated changes — easier to review and merge independently.

phillipc:
@brianmhunt Sure, done, sometimes `git revert` actually works ;)

brianmhunt:
@phillipc This changes the default behavior for `@tko/build.reference` users. Previously, `==` was silently overloaded to `===` at startup via `overloadOperator`. With this PR, `options.overloadEvilTwins` defaults to `false`, so `==` reverts to loose equality.

That's a breaking change for every existing `@tko/build.reference` binding that uses `==` — they've been getting strict equality for years.

The default should be `true` (preserving existing behavior), with the option to explicitly set `false` for loose equality. Could you flip the default?

brianmhunt:
Update on this — I pushed a fix to the default but then realized the bigger issue: the `if (options.overloadEvilTwins)` check in the operator functions runs on every `==`/`!=` evaluation in every binding expression. That's the parser's hot path.

The original `overloadOperator` pattern was actually the right approach — mutate the operator table once at startup, zero per-call overhead. I've reverted my changes on the branch.

I think we should close this PR and keep the existing `overloadOperator` API. The current code works correctly and is performant. If we want to make the evil-twin behavior configurable, we can add an `options.overloadEvilTwins` flag that calls `overloadOperator` at init time (one-time cost) rather than checking a flag per evaluation.

@phillipc what do you think?

phillipc:
@brianmhunt I know but we have no init-hook. We would then also have to differentiate between dynamic runtime options and static init options. That would certainly be the best way forward, but more extensive and should be knockout compatible designed.

brianmhunt:
## Alternative approach: PR #314

We've opened #314 with a different approach that avoids adding a branch to the hot-path operator functions.

### `ko.options.strictEquality`

Instead of checking `options.overloadEvilTwins` on every `==`/`!=` evaluation, #314 swaps the function references in the operators table at configuration time — same mechanism as the existing `overloadOperator`, but as a clean setter on `ko.options`:

```js
ko.options.strictEquality = true   // == becomes ===, != becomes !==
```

Zero per-evaluation cost. Strongly typed. Fully tested (6 new tests).

Also introduces `defineOption()` — a plugin-friendly API for registering custom options with side-effect setters, so other packages can extend `ko.options` the same way.

See #314 for details.

PR #294, opened 2026-04-10
fix: correct propertyChangeFired typo

phillipc:
This has always been there, and with the TypeScript port, it was also made into a property. Never set `true` and not detected because a IE10 to properly 11 case.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue with the IE autocomplete workaround where the blur event handler was referencing incorrect internal state, ensuring proper behavior during autocomplete interactions.


brianmhunt (approved):
Retro-reviewed. Legitimate bugfix — corrects propertyChangeFired typo to propertyChangedFired in value.ts IE autocomplete handler.

PR #295, opened 2026-04-12
Fixing package.json module paths and remove helpers

phillipc:
- Fixing  the repackage script to conditionally include "helpers/" based on its existence.
- Fix the "module" entry in multiple package.json files to point to "dist/index.js" instead of specific binding files.

## Summary by CodeRabbit

* **Chores**
  * Standardized package ESM entry points to a single index file for all packages.
  * Simplified package exports by removing secondary ./helpers/* subpath mappings.
  * Reduced published package footprint by removing the helpers directory from distributions.
  * Packaging now conditionally includes helpers only when that directory exists.

### DIFF
--- a/builds/reference/package.json
+++ b/builds/reference/package.json
@@ -34,7 +34,6 @@
   },
   "files": [
     "dist/",
-    "helpers/",
     "types/"
   ],
   "homepage": "https://tko.io",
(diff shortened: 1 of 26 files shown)


brianmhunt:
Superseded by #308 which applies these fixes (module paths + helpers removal + repackage.mjs deletion) cleanly to current main. Thanks for identifying the issue @phillipc.

PR #312, opened 2026-04-15
Fix knip findings: unused files, deps, unlisted deps

brianmhunt:
## Summary

Safe fixes for knip findings — no public API changes.

- Delete 3 unused files (CJS shims, package template)
- Remove 8 unused `@tko/*` dependencies from package.json
- Add 10 unlisted `@tko/*` dependencies to package.json
- Clean up knip config hints
- Add `bun run verify` script (biome + tsc + build + test)
- Add safe-change workflow guidance to AGENTS.md

## Remaining knip findings (needs public API review)

| Category | Count | Notes |
|---|---|---|
| Unused exports | 10 | Functions/classes — may be used by external consumers |
| Unused exported types | 19 | Type-only — may be used by downstream packages |
| Duplicate exports | 2 | Aliases (harmless) |

These will be addressed after reviewing downstream consumer impact.

## Test plan

- [x] `bun run verify` passes (biome + tsc + build + 2679 tests)
- [x] Each commit verified independently

### DIFF
--- a/builds/knockout/package.json
+++ b/builds/knockout/package.json
@@ -43,14 +43,14 @@
     "@tko/builder": "^4.0.1",
     "@tko/filter.punches": "^4.0.1",
     "@tko/provider.attr": "^4.0.1",
-    "@tko/provider.bindingstring": "^4.0.1",
     "@tko/provider.component": "^4.0.1",
     "@tko/provider.databind": "^4.0.1",
     "@tko/provider.multi": "^4.0.1",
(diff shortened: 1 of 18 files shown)


phillipc:
@brianmhunt "verbatimModuleSyntax": true is missing. In this Way, we can require "import type" usage

phillipc:
tslib is also unneeded, we can set importHelpers to false and downlevelIteration, too. https://www.typescriptlang.org/tsconfig/#importHelpers

PR #314, opened 2026-04-15
Replace overloadOperator with zero-cost setStrictEquality

### LINKED ISSUE #290: Error parsing Binding expressions with Operators == and != in reference Build

brianmhunt:
## Summary

Replaces `overloadOperator()` with `setStrictEquality(bool)` — a configuration-time function swap that has **zero runtime cost** in the expression evaluation hot path.

## Problem

PR #292 proposed adding `if (options.overloadEvilTwins)` branches inside the `==` and `!=` operator functions. These operators are called during every binding expression evaluation — a high-frequency inner loop. Adding a branch there degrades performance for all users.

The existing `overloadOperator()` API is better (it swaps function references), but it's overly general for a feature only used for `==`/`!=` → `===`/`!==`.

## Solution

`setStrictEquality(strict: boolean)` swaps pre-built function references in the operators table. Each replacement function has the correct `.precedence` metadata so the shunting-yard parser works correctly.

- `setStrictEquality(true)` → `==` becomes `===`, `!=` becomes `!==`
- `setStrictEquality(false)` → restores default loose equality
- `===` and `!==` operators are never affected

`@tko/build.reference` calls `setStrictEquality(true)` at setup (preserving current behavior). `@tko/build.knockout` uses the default (loose equality).

## Changes

- `packages/utils.parser/src/operators.ts` — add `setStrictEquality`, extract loose/strict equality functions with precedence
- `packages/utils.parser/src/index.ts` — export `setStrictEquality`, remove `overloadOperator`
- `builds/reference/src/index.ts` — use `setStrictEquality(true)` instead of `overloadOperator`
- `packages/utils.parser/spec/strictEqualityBehaviors.ts` — 6 new tests

## Test plan

- [x] 6 new tests: loose default, strict mode, inequality, toggle, === isolation
- [x] All 2685 tests pass (2679 existing + 6 new)
- [x] `bunx tsc` — 0 errors
- [x] `bunx @biomejs/biome check .` — clean

Closes #290. Alternative to #292.

## Summary by CodeRabbit

* **New Features**
  * Added a global option to toggle strict vs. loose equality for `==`/`!=` in binding expressions.

* **Breaking Changes**
  * Removed the low-level operator customization API; equality behavior is now controlled via the new configuration option.

* **Tests**
  * Added comprehensive tests covering equality semantics, mixing with logical operators, and option toggling.

### DIFF
--- a/builds/reference/src/index.ts
+++ b/builds/reference/src/index.ts
@@ -20,13 +20,12 @@ import components from '@tko/utils.component'
 import { createElement, Fragment } from '@tko/utils.jsx'
 import { JsxObserver } from '@tko/utils.jsx'
 
-import { overloadOperator } from '@tko/utils.parser'
+import { options } from '@tko/utils'
 
 declare const BUILD_VERSION: string
 
-/** Overload "evil twins" with strict equivalents */
-overloadOperator('==', (a, b) => a === b)
-overloadOperator('!=', (a, b) => a !== b)
+/** Use === and !== instead of == and != in binding expressions */
+options.strictEquality = true
 
 const builder = new Builder({
   filters,
--- a/packages/utils.parser/spec/strictEqualityBehaviors.ts
+++ b/packages/utils.parser/spec/strictEqualityBehaviors.ts
@@ -1,0 +1,73 @@
+import { expect } from 'chai'
+
+import { options } from '@tko/utils'
+import { Parser } from '../dist'
+
+function ctxStub(ctx?) {
+  return {
+    lookup(v) {
+      return ctx?.[v]
+    },
+    $data: ctx
+  }
+}
+
+function evaluate(expr: string, vars: object = {}) {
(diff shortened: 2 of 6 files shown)


brianmhunt:
@coderabbitai pls review

brianmhunt:
@coderabbitai pls review

phillipc:
Good Compromise!

PR #328, opened 2026-04-16
ComponentABC: auto-discover <template id> for multi-instance reuse

brianmhunt:
## Summary

**Stacks on #326.** Combines PR 326's children-as-template (single-instance, inline) with multi-instance reuse via a simple convention: `<template id="\$customElementName">` is auto-discovered.

```html
<template id="hello-world">
  <h1 data-bind="text: name"></h1>
</template>

<hello-world></hello-world>
<hello-world></hello-world>
<hello-world></hello-world>

<script>
class HelloWorld extends ko.Component {
  name = ko.observable('world')
}
HelloWorld.register()  // no template config, no element override
</script>
```

All three `<hello-world>` instances share the template by id. No boilerplate. Explicit `static get template`/`static get element` still works when you want to override.

## Resolution order in ComponentABC.template

1. `'template' in prototype` → instance hook
2. Subclass overrides `static get template()` → use that
3. Subclass overrides `static get element()` → `{ element: this.element }`
4. **NEW**: `document.getElementById(customElementName)` exists → use it
5. Otherwise undefined → children-as-template (from #326)

## Why this is the right marriage

The user asked: "how do we marry using one's own children with re-use of those children in multiple components?" Three options considered:

- **A**: Status quo (`static get template() { return { element: 'hello-world' } }`) — works today, but boilerplate.
- **B (this PR)**: Auto-discover by name. Zero boilerplate for the common case. Falls back to #326's children-as-template.
- **C**: First-instance-donates — magical, timing-sensitive, hard to explain. Rejected.

## Tests

Two new tests in `ComponentABCBehaviors`:

- `auto-discovers <template id="\$customElementName"> when no template or element is set` — two instances render identically from the shared template.
- `falls back to children-as-template when no matching <template id> exists` — inline markup still works per-instance.

Full suite: 2700 passed, 42 skipped, 0 failed.

## Test plan

- [x] `bun run test` green
- [x] Existing ComponentABC tests (explicit template, explicit element, class-kebab-case derivation, dispose) unchanged

### DIFF
--- a/packages/binding.component/src/componentBinding.ts
+++ b/packages/binding.component/src/componentBinding.ts
@@ -43,8 +43,7 @@ export default class ComponentBinding extends DescendantBindingHandler {
     if (typeof document === 'undefined') return null
     const el = document.getElementById(componentName)
     if (!el) return null
-    const source =
-      (el as any).content && (el as any).content.nodeType === 11 ? (el as any).content.childNodes : el.childNodes
+    const source = el instanceof HTMLTemplateElement ? el.content.childNodes : el.childNodes
     return makeArray(source)
   }
 
--- a/packages/binding.component/src/componentBinding.ts
+++ b/packages/binding.component/src/componentBinding.ts
@@ -54,9 +54,13 @@ export default class ComponentBinding extends DescendantBindingHandler {
    * as "no children" so `<my-comp>   </my-comp>` still errors.
    */
   hasMeaningfulChildren(): boolean {
-    return this.originalChildNodes.some(
-      n => n.nodeType === Node.ELEMENT_NODE || (n.nodeType === Node.TEXT_NODE && (n.nodeValue ?? '').trim().length > 0)
-    )
+    const nodes = this.originalChildNodes
+    for (let i = 0; i < nodes.length; i++) {
+      const n = nodes[i]
+      if (n.nodeType === Node.ELEMENT_NODE) return true
+      if (n.nodeType === Node.TEXT_NODE && (n.nodeValue ?? '').trim().length > 0) return true
+    }
+    return false
   }
 
   cloneTemplateIntoElement(componentName: string, template: any, element: Node) {
--- a/packages/binding.component/src/componentBinding.ts
+++ b/packages/binding.component/src/componentBinding.ts
@@ -34,6 +34,21 @@ export default class ComponentBinding extends DescendantBindingHandler {
   }
 
   /**
+   * Look up a <template id="$componentName"> (or any element with that id)
+   * in the document and return its child nodes as a template node array,
+   * or null if none is found. Used as a fallback when no config/viewModel
+   * template and no instance children are provided.
+   */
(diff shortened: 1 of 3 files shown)


packages/binding.component/src/componentBinding.ts
```diff
@@ -33,15 +33,33 @@ export default class ComponentBinding extends DescendantBindingHandler {
...
    * True when originalChildNodes contain at least one element or a text
    * node with non-whitespace content. Whitespace-only children are treated
    * as "no children" so `<my-comp>   </my-comp>` still errors.
    */
   hasMeaningfulChildren(): boolean {
-    return this.originalChildNodes.some(
-      n => n.nodeType === Node.ELEMENT_NODE || (n.nodeType === Node.TEXT_NODE && (n.nodeValue ?? '').trim().length > 0)
-    )
+    const nodes = this.originalChildNodes
+    for (let i = 0; i < nodes.length; i++) {
```
brianmhunt, line 57:
```
const nodes = this.originalChildNodes;
// 1. Cache the length
for (let i = 0, len = nodes.length; i < len; i++) {
  const n = nodes[i];
  
  if (n.nodeType === Node.ELEMENT_NODE) return true;
  
  // 2 & 3. Drop '??' and use Regex to avoid string allocation
  if (n.nodeType === Node.TEXT_NODE && /\S/.test(n.nodeValue)) {
    return true;
  }
}
```

brianmhunt: or

```
for (const n of this.originalChildNodes) {
  if (n.nodeType === Node.ELEMENT_NODE) return true;
  if (n.nodeType === Node.TEXT_NODE && /\S/.test(n.nodeValue)) return true;
}
```

brianmhunt:
We should define an option for the 

```ts
componentConventionalTemplate = (component?: c) => string
// e.g.
componentConventionalTemplate = (c) => `ko-${c.elementName}`
```

packages/binding.component/src/componentBinding.ts
```diff
@@ -33,15 +33,33 @@ export default class ComponentBinding extends DescendantBindingHandler {
...
 
+  /**
+   * Look up a <template id="$componentName"> (or any element with that id)
+   * in the document and return its child nodes as a template node array,
+   * or null if none is found. Used as a fallback when no config/viewModel
+   * template and no instance children are provided.
+   */
+  findSharedTemplate(componentName: string): Node[] | null {
+    if (typeof document === 'undefined') return null
+    const el = document.getElementById(componentName)
```
brianmhunt, line 44: just use document?.getElementByID; no typeof

packages/binding.component/src/componentBinding.ts
phillipc: - `if` always with `{ }` (But tokens cost money) 
- `of` is modern 
- Drop `??` is good, but perfectly fine for simple things

PR #330, opened 2026-04-17
Split vitest into unit (happy-dom) + browser projects

brianmhunt:
## Summary

Most package specs don't need a real browser. Split them into a fast **happy-dom** project (~5× faster startup) and keep a real-browser project for specs that depend on MutationObserver timing, form-control event ordering, JSX observer lifecycles, async DOM reflow, or the built bundles.

## Numbers

| project | files | tests | duration |
|---|---|---|---|
| unit (happy-dom) | 39 | 766 | ~1.1s |
| browser (playwright) | 105 | 1942 | ~6.7s |

`bun run test` runs both concurrently — wall time remains browser-bound (~7s). The win is **local fast-feedback iteration**:

```bash
bun run test:unit      # ~1s, no playwright/build required
bun run test:browser   # full browser matrix
```

## Browser-project allow-list

Anything else is happy-dom by default:

```
packages/bind
packages/binding.component
packages/binding.core
packages/binding.foreach
packages/binding.if
packages/binding.template
packages/builder
packages/provider.component
packages/utils.component
packages/utils.jsx
packages/utils/spec/utilsDomBehaviors.ts
```

## Notes

- Declared `happy-dom` as a direct devDependency (previously pulled transitively via vitest's optional peer).
- Total test count shifts from 2701 → 2698 passed. 3 tests depended on quirky real-browser behaviors and appear to be silently skipped by happy-dom. No failures, no new tests flagged as skipped. Could investigate further as a follow-up if those tests turn out to be load-bearing.
- CI workflow unchanged — each browser matrix job still runs both projects. Splitting the unit portion out of the browser matrix is a possible follow-up but adds CI complexity for marginal gain.

## Test plan

- [x] `bun run test` — 2698 pass, 0 fail, 42 skip (all previously skipped)
- [x] `bun run test:unit` — 766 tests in ~1s
- [x] `bun run test:browser` — 1942 tests in ~7s

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,8 @@
   "scripts": {
     "build": "bun run --filter './packages/*' build && bun run --filter './builds/*' build",
     "test": "bunx @biomejs/biome ci . && bun run build && bunx vitest run",
+    "test:unit": "bunx vitest run --project unit",
+    "test:browser": "bun run build && bunx vitest run --project browser",
     "test:ff": "bunx @biomejs/biome ci . && bun run build && VITEST_BROWSERS=firefox bunx vitest run",
     "tsc": "bunx tsc",
     "dts": "bunx tsc --build tsconfig.dts.json",
@@ -38,6 +40,7 @@
     "@vitest/browser-playwright": "^4.1.4",
     "chai": "^6.2.2",
     "esbuild": "^0.28.0",
+    "happy-dom": "^20.9.0",
     "jquery": "^4.0.0",
     "knip": "^6.4.1",
     "playwright": "^1.59.1",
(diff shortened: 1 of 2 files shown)


brianmhunt:
Closing — on reflection this trades coverage for speed in a way that doesn't fit a low-level DOM binding library aimed at broad audiences.

The 9 files excluded from happy-dom (form-control event ordering, MutationObserver, custom elements, namespaced attributes) are exactly the behaviors TKO's value proposition rests on and where real browsers disagree with each other and with partial-DOM implementations. Skipping those in happy-dom concentrates the riskiest tests in a single environment rather than expanding coverage.

happy-dom's partial DOM is a correctness hazard for a DOM binding library, not just a speed tradeoff — a false green there would be worse than no signal at all. And the wall-time win is marginal (browser-bound at ~7s combined) while the real local iteration speed already comes from targeted package runs (e.g. `bunx vitest run packages/observable` ≈ 1s warm).

Filing guidance in AGENTS.md so future agents don't re-propose the same trade.

PR #331, opened 2026-04-17
Document TKO's testing coverage bar (real browsers only)

brianmhunt:
## Summary

Codifies the coverage policy for TKO in AGENTS.md, under the existing **Testing** section. Motivated by the closed [PR #330](https://github.com/knockout/tko/pull/330) — I proposed a happy-dom split for speed, and the trade didn't hold up for a library at TKO's level. Filing the reasoning so future agents don't re-propose the same pattern.

## The policy

TKO is a low-level DOM binding framework published to npm and used by unknown downstream apps. The dark-factory thesis (small teams + AI assistance maintaining a framework) puts the test suite in the role humans used to play. So the coverage bar is higher than typical.

Three concrete rules:

1. **Every spec runs in every supported real browser.** CI matrix (chromium, firefox, webkit) is load-bearing.
2. **No partial-DOM environments.** happy-dom/jsdom/linkedom give false signal for a DOM binding library — the behaviors TKO most needs to verify (form-control event ordering, MutationObserver, namespaced attributes, custom elements) are exactly where partial DOMs diverge.
3. **Excluding files from a run is a red flag.** Concentrates risk; doesn't save coverage.

Also points at PR #330 as a worked example so the rationale is traceable.

## Test plan

Docs-only. No code changes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -59,8 +59,7 @@ Individual packages can be built from their directory with `bun run build`.
 - **Test files**: `packages/*/spec/**/*.ts`, `builds/*/spec/**/*.js`
 - **Run**: `bunx vitest run` (all tests) or `bunx vitest run <path>` (single file)
 
-Tests run in a real browser via Playwright — not jsdom. This is required
-because TKO does low-level DOM manipulation, MutationObserver, and event handling.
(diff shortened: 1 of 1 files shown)


brianmhunt:
Closing — I misunderstood the intent. The goal of happy-dom/node isn't to replace browser testing for speed, it's to add coverage for non-browser runtimes TKO should support (TUI, SSR, server-side reactivity). Will redo as additive coverage, not a swap. Leaving the closed-PR reference intact for context.

PR #345, opened 2026-04-20
Fix ?? (nullish coalescing) earlyOut to not short-circuit on falsy values

phillipc:
## Summary

The `??` operator's `earlyOut` was `a => a`, identical to `||`. This caused `??` to short-circuit on falsy values like `0`, `''`, and `false`, which is incorrect — `??` should only short-circuit when the value is not `null` and not `undefined`.

## Fix

```diff
-operators['??'].earlyOut = a => a
+operators['??'].earlyOut = a => a !== null && a !== undefined
```

## Verification

All 404 parser tests pass. Full suite (5371 tests) passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed nullish coalescing operator behavior to correctly handle `null` and `undefined` values separately from other falsy values.

### DIFF
--- a/packages/utils.parser/spec/nodeBehaviors.ts
+++ b/packages/utils.parser/spec/nodeBehaviors.ts
@@ -54,7 +54,10 @@ describe('Operators', function () {
 
   it('?? evaluates rhs iff lhs is null/undefined', () => {
     let rhsCalls = 0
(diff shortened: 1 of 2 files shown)


brianmhunt (approved):
👍 test added

PR #346, opened 2026-04-20
Remove dead TextInputLegacyFirefox class

phillipc:
## Summary

Removes the `TextInputLegacyFirefox` class from `packages/binding.core/src/textInput.ts`.

## Why

The class was dead code — it overrode a method named `eventsIndicatingValueChange()` which doesn't exist on the parent. The actual method is `eventsIndicatingSyncValueChange()`, so the override never took effect. Additionally, Firefox 3.x (the target) has been EOL for over a decade.

## Verification

All 46 textInput tests pass.


brianmhunt (approved):
👍 looks good

PR #347, opened 2026-04-20
Remove duplicate options import in computed.ts

phillipc:
## Summary

Removes the duplicate `options` import from `packages/computed/src/computed.ts`. The module was importing both `options` and `options as koOptions` from `@tko/utils`. The bare `options` reference at line 430 is replaced with the existing `koOptions` alias.

## Verification

Full test suite passes (5371 tests).

### DIFF
--- a/packages/computed/src/computed.ts
+++ b/packages/computed/src/computed.ts
@@ -8,7 +8,6 @@ import {
   arrayForEach,
   domNodeIsAttachedToDocument,
   extend,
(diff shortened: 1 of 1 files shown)


phillipc:
The reason for using the koOptions alias is name conflicts with ComputedOptions.

brianmhunt (approved):
👍  Thanks @phillipc

PR #350, opened 2026-04-21
Add TypeScript code review skill and updated findings plans

phillipc:
## Summary

Adds the TypeScript code review skill and four rounds of code review findings plans, updated to reflect the current TKO toolchain and review feedback.

**Supersedes:** #297 (closed draft)

## What changed since #297

- Incorporated review feedback from @brianmhunt on #297
- Updated all findings plans (rounds 1–4) to current stack baseline (Bun + Biome + Vitest)
- Deduplicated findings across rounds — each finding now has exactly one canonical location
- Round 1 archived (historical context only, no active findings)
- Added verification status table in Round 2 tracking Brian's five critical spot-check findings:
  - Proxy `deleteProperty` trap — **fixed** in #336
  - `??` earlyOut identical to `||` — **confirmed**, PR #345 open
  - Parser operator-precedence — Issue #342 (closed)
  - `TextInputLegacyFirefox` dead code — **confirmed**, PR #346 open
  - `style` binding jQuery reference — PR #339 (merged)
- Round 3: Added independent verification notes for the `??` parser finding
- Round 4: Cross-referenced the `deleteProperty` fix (#336) as validation for the related `getOwnPropertyDescriptor` finding
- Added PR #297 review notes section in Round 4
- Removed stale tooling references (Karma, Make, lerna) throughout
- Skill document (`skills/typescript-code-review/`) and `tsconfig.json` wiring included

## Files changed

- `skills/typescript-code-review/SKILL.md` — code review skill definition
- `skills/typescript-code-review/references/tko-conventions.md` — TKO-specific conventions
- `plans/typescript-code-review-findings.md` — Round 1 (archived)
- `plans/typescript-code-review-findings-2.md` — Round 2 (active backlog + verification status)
- `plans/typescript-code-review-findings-3.md` — Round 3 (active backlog)
- `plans/typescript-code-review-findings-4.md` — Round 4 (active backlog)
- `tsconfig.json` — exclude `skills/` directory

## Verification

- `bun run tsc` — no type errors introduced (plans/skills are .md files)
- `bun run check` — lint/format clean
- All findings cross-checked against current `main` branch

## Summary by CodeRabbit

* **Documentation**
  * Added an AI Skills section with a new TypeScript code-review skill, including structured review workflow, output format, and extensive reference & convention guidance.
  * Added multi-round TypeScript review plans (Rounds 1–5) capturing critical/important/suggestion findings, verification checklists, de-duplication rules, and remediation backlogs.

* **Chores**
[...]

PR #352, opened 2026-04-21
Fix makeTemplateNode bypassing HTML sanitization

phillipc:
## Problem

`makeTemplateNode` in `packages/binding.foreach/src/foreach.ts` assigned `sourceNode.text` directly to `parentNode.innerHTML` for `<script>` template elements, bypassing all three HTML sanitization controls:

- `options.templateSizeLimit`
- `options.allowScriptTagsInTemplates`
- `options.sanitizeHtmlTemplate`

- Detected in https://github.com/knockout/tko/pull/350

## Fix

- Export `validateHTMLInput` from `@tko/utils` (was previously internal to `html.ts`)
- Route the `<script>` element's `.text` through `validateHTMLInput()` before assigning to `innerHTML`

This ensures the same size-limit, script-tag, and custom sanitizer checks apply to foreach `<script>` templates as they do everywhere else via `parseHtmlFragment`.

## Testing

All 158 existing foreach tests pass (happy-dom + chromium).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed HTML sanitization being bypassed for script-based foreach templates. Templates now properly enforce size limits, nested script detection, and HTML sanitization checks. Applications that embed script markup in foreach templates must enable `allowScriptTagsInTemplates: true`.

* **Tests**
  * Added comprehensive test coverage for foreach script template sanitization, including template size limit validation and nested script markup detection.

### DIFF
--- a/.changeset/fix-foreach-script-template-sanitization.md
+++ b/.changeset/fix-foreach-script-template-sanitization.md
@@ -1,0 +1,8 @@
+---
+'@tko/binding.foreach': patch
+---
+
+Fix `makeTemplateNode` bypassing HTML sanitization for `<script>` foreach templates. Script-template text now
+goes through the existing HTML template parsing safeguards, so oversized templates, disallowed nested `<script>`
+markup, and `sanitizeHtmlTemplate` apply consistently. Existing apps that intentionally embed `<script>` markup
+in foreach script templates should set `allowScriptTagsInTemplates: true`.
--- a/packages/binding.foreach/spec/eachBehavior.ts
+++ b/packages/binding.foreach/spec/eachBehavior.ts
@@ -176,6 +176,67 @@ describe('each binding', function () {
     $template.remove()
   })
 
+  it('applies the template size limit to named <script> templates', function () {
+    const originalTemplateSizeLimit = options.templateSizeLimit
+    const target = $('<ul data-bind=\'foreach: {name: "tID", data: $data}\'>Zee</ul>')
+    const $template = $("<script type='text/ko-template' id='tID'></script>").appendTo(document.body)
+    $template.text('<li data-bind="text: $data"></li>')
+
+    try {
+      options.templateSizeLimit = 10
+
+      assert.throws(function () {
+        applyBindings(['G1'], target[0])
+      }, /Template is too long/)
+    } finally {
+      options.templateSizeLimit = originalTemplateSizeLimit
+      $template.remove()
+    }
+  })
+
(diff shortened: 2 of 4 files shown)


brianmhunt:
## Adversarial review

Fix at `foreach.ts:60` lands at the right site. Bypass is real. Verified no other unsanitized HTML sinks in `packages/**/src`.

**Must fix before merge:**
1. **Regression test** — none added. Per AGENTS.md ("missing test that would have caught the bug"). Need: `templateSizeLimit` throw, `allowScriptTagsInTemplates: false` throw, `sanitizeHtmlTemplate` invocation.
2. **Changeset** — none in diff. Post-#377 release pipeline won't bump versions without one. Both `@tko/binding.foreach` and `@tko/utils` change behavior.

**Worth deciding:**
3. **Refactor to `parseHtmlFragment`** instead of exporting `validateHTMLInput`. `parseHtmlFragment` + `parseHtmlForTemplateNodes` already do the validate-then-parse contract; matches how `binding.template` and `utils.component` consume HTML strings. Avoids adding `validateHTMLInput` to `@tko/utils` public API as a SemVer commitment.
4. **Behavior change is breaking-ish.** Silent-accept → throw on existing apps with `<script>` foreach templates containing `<script>` bodies under `allowScriptTagsInTemplates: false`. Changeset body should call out migration: set `allowScriptTagsInTemplates: true` if intentional.

**Out of scope (note for the record):**
- `<template>` and `<div>` branches in `makeTemplateNode` don't run TKO sanitization — they arrive as already-parsed DOM, not strings. TKO controls are string-only by design. Confirm this matches threat model.

brianmhunt:
**Changeset how-to** (sorry, should have included this):

```bash
bunx changeset add
# select @tko/binding.foreach AND @tko/utils, pick "patch" (or "minor" if treating the silent→throw shift as breaking-ish), paste description
# commit the generated .changeset/<slug>.md with the fix
```

Or hand-write `.changeset/fix-foreach-sanitize-script-template.md`:

```md
---
'@tko/binding.foreach': patch
'@tko/utils': patch
---

Fix `makeTemplateNode` bypassing HTML sanitization for `<script>` foreach templates. `validateHTMLInput` is now exported from `@tko/utils`. Behavior change: previously silent acceptance of oversized templates / embedded `<script>` bodies under `allowScriptTagsInTemplates: false` now throws. If you intentionally embed `<script>` content in a foreach template, set `allowScriptTagsInTemplates: true`.
```

phillipc:
@brianmhunt I have implemented the requested changes

PR #362, opened 2026-04-22
Fix Parser instantiation and globals in ComponentProvider

phillipc:
## Summary

`ComponentProvider.getComponentParams` was instantiating `Parser` with a downcast (`new (Parser as any)(node, context, this.globals)`) and passing `undefined` as globals to `.parse()`. Since `Parser` takes no constructor arguments, the globals were silently dropped — global variables could never resolve in component params.

## Changes

- **`ComponentProvider.ts`**: Replace `new (Parser as any)(node, context, this.globals) as Parser` with `new Parser()`, and pass `this.globals` to `.parse()` instead of `undefined`. This matches how every other provider (`BindingStringProvider`, `mustacheParser`) uses `Parser`.
- **`componentProviderBehaviors.ts`**: Add test that registers a component with `params="answer: GLOBAL_CONST"` where `GLOBAL_CONST` is provided via provider globals, verifying globals resolution in component params.

## Verification

All 52 existing tests pass (chromium + happy-dom). New test passes in both environments.

### DIFF
--- a/.changeset/fix-component-provider-parser-globals.md
+++ b/.changeset/fix-component-provider-parser-globals.md
@@ -1,0 +1,10 @@
+---
+"@tko/provider.component": patch
+---
+
+Fix component params globals resolution in `ComponentProvider`.
+
+`getComponentParams` now instantiates `Parser` with `new Parser()` and passes
+provider globals to `parser.parse(...)`, matching canonical parser usage in
+other providers. This restores resolution of globals in component `params`
+expressions (for example, `params="answer: GLOBAL_CONST"`).
--- a/packages/provider.component/spec/componentProviderBehaviors.ts
+++ b/packages/provider.component/spec/componentProviderBehaviors.ts
@@ -150,6 +150,29 @@ describe('Components: Provider', function () {
       applyBindings({}, ne)
       // No error raised.
     })
+
+    it('resolves globals in component params', function () {
+      const provider = new MultiProvider({
+        providers: [new DataBindProvider(), new ComponentProvider()],
+        globals: { GLOBAL_CONST: 42 }
+      })
+      options.bindingProviderInstance = provider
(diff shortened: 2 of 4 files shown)


brianmhunt:
## Adversarial review

Fix matches canonical usage in `BindingStringProvider.ts:53` and `mustacheParser.ts:20`. New test would have failed pre-fix (globals defaulted to `{}`, `GLOBAL_CONST` resolves to undefined, `expect(42)` fails).

**Must fix before merge:**
1. **Changeset** — none in diff. `@tko/provider.component` is a user-visible bug fix and the post-#377 release pipeline won't bump versions without one.

**Worth fixing while here:**
2. **Stale Parser JSDoc** — `packages/utils.parser/src/Parser.ts:34-40` documents `new Parser(node, context)` with `node`/`context`/`globals` parameters. The class has no constructor; the doc has been a lie since the constructor was removed and is exactly what tricked the original ComponentProvider author into the `as any` downcast. Cheap to delete the parameter list while you're touching this surface.

**Out of scope (note for the record):**
- Test doesn't unregister the `xenon` component or restore `options.bindingProviderInstance` — consistent with every sibling test in this file. Pre-existing pattern.

brianmhunt:
**Changeset how-to** (sorry, should have included this):

```bash
bunx changeset add
# select @tko/provider.component, pick "patch", paste short description
# commit the generated .changeset/<slug>.md with the fix
```

Or hand-write `.changeset/fix-component-provider-globals.md`:

```md
---
'@tko/provider.component': patch
---

Fix Parser instantiation and globals in ComponentProvider — globals were silently dropped from component params resolution.
```

phillipc:
@brianmhunt done

phillipc:
@brianmhunt, are you living this project? Please give me the right to merge, to keep TKO alive, thanks.

PR #363, opened 2026-04-22
Fix removeEventListener missing options in LifeCycle

phillipc:
## Problem

`LifeCycle.__addEventListener` registers listeners with `addEventListener(type, handler, options)` but disposes them with `removeEventListener(type, handler)` — without passing `options`.

Per the DOM spec, `removeEventListener` must match the `capture` flag used during registration. Without it, capture-phase listeners remain attached after disposal, leaking event handlers.

## Fix

Pass `options` through to `removeEventListener` in the dispose callback.

## Test

Added a test that registers a capture-phase listener, verifies it fires, disposes the LifeCycle, and confirms the listener no longer fires.


brianmhunt (approved):
👍 Thanks @phillipc

PR #369, opened 2026-04-24
fix(binding.foreach): add afterRender/beforeMove/afterMove callbacks (#329)

### LINKED ISSUE #329: ForEachBinding does not match the documented callback API of foreach
In build.reference, the `foreach` Binding behavior comes from `ForEachBinding`. However, the `foreach` Binding documentation describes the callback behavior provided by `TemplateForEachBindingHandler` used in build.knockout, resulting the documented API and the actual runtime behavior in the reference build to not fully match.

### Problematic behavior
`ForEachBinding` is only partially compatible with the [callback API](https://www.tko.io/bindings/foreach-binding/#note-8-post-processing-or-animating-the-generated-dom-elements) described for the `foreach` Binding:

- `afterRender `is not supported by `ForEachBinding`
- `beforeMove` is not supported by `ForEachBinding`
- `afterMove` is not supported by `ForEachBinding`
- `afterAdd` exists in `ForEachBinding`, but uses a different callback contract than `TemplateForEachBindingHandler`
- `beforeRemove` exists in `ForEachBinding`, but also uses a different callback contract than `TemplateForEachBindingHandler`

### Expected behavior
The behavior of `foreach` in build.reference should be aligned with one of these expectations:

- `ForEachBinding` should support the same documented callback API as `TemplateForEachBindingHandler`, or
- the `foreach` Binding documentation should clearly distinguish between the callback behavior of `ForEachBinding` and `TemplateForEachBindingHandler`

### Why this is a problem
This creates a silent compatibility gap between the documented `foreach` API and the actual behavior of `ForEachBinding` in the reference build. Rendering may still work, but lifecycle hooks such as `afterRender` do not run, and callbacks such as `afterAdd` and `beforeRemove` are invoked with a different signature than the one described in the documentation.

### Suggested clarification
Please clarify whether the different callback behavior of `ForEachBinding` is intentional. If it is intentional, the documentation should explicitly document the differences from `TemplateForEachBindingHandler`. If it is not intentional, `ForEachBinding` should be brought closer to the documented `foreach` callback behavior.

phillipc:
## Summary

Closes #329.

Adds the documented `foreach` lifecycle callbacks that were missing from `ForEachBinding` in the reference build, without breaking the existing `afterAdd`/`beforeRemove` batch contract:

- **`afterRender(nodes, value)`** — fires once per item when a new nodeset is rendered from the template (initial render and later additions).
- **`beforeMove(node, newIndex, value)`** / **`afterMove(node, newIndex, value)`** — fire per retained node whose index changed; move detection compares the tracked `firstLastNodesList` entries against the current array values and only runs if at least one move callback is registered.
- **`afterAdd({ nodeOrArrayInserted, foreachInstance })`** and **`beforeRemove({ nodesToRemove, foreachInstance })`** are unchanged, including the tko-specific thenable-return shortcut for deferred node removal.
- Documentation (`foreach-binding.md` Note 8) now calls out the reference build's batch-shaped `afterAdd`/`beforeRemove` and the thenable extension as deviations from the original Knockout contract.

## Test plan

- [x] `bun run verify` (5395 tests pass, 0 failures)
- [x] New spec coverage:
  - `afterRender` fires for initial and added items with `(nodes, value)`
  - `beforeMove`/`afterMove` fire with `(node, newIndex, value)` when retained items shift
- [x] Existing `afterAdd`/`beforeRemove` tests continue to pass unchanged (batch signature + thenable removal)
- [x] `verified-behaviors.json` updated
- [x] Changeset added (`patch`, additive only)

## Summary by CodeRabbit

* **New Features**
  * Foreach binding adds per-item lifecycle callbacks: afterRender (newly rendered items), beforeMove and afterMove (items that change position).

* **Documentation**
  * Foreach docs updated with per-item callback signatures, compatibility notes, and that beforeRemove may return a thenable (tko-specific).

* **Tests**
  * New tests cover afterRender, beforeMove, afterMove, move detection, and non-reactivity of observables read during afterRender.

### DIFF
--- a/.changeset/fix-foreach-callback-api.md
+++ b/.changeset/fix-foreach-callback-api.md
@@ -1,0 +1,9 @@
+---
+"@tko/binding.foreach": patch
+---
+
+Add documented `foreach` lifecycle callbacks to `ForEachBinding`:
+`afterRender(nodes, value)` fires when an item is rendered from the template,
+`beforeMove(node, newIndex, value)` and `afterMove(node, newIndex, value)` fire
+for retained items that shift position. The existing `afterAdd` and
+`beforeRemove` signatures are unchanged.
--- a/packages/binding.foreach/spec/eachBehavior.ts
+++ b/packages/binding.foreach/spec/eachBehavior.ts
@@ -22,7 +22,7 @@ import { bindings as coreBindings } from '@tko/binding.core'
 
 import { ForEachBinding } from '../dist/foreach'
 
-import $ from 'jquery'
+import $ from 'jquery' //TODO Should be removed and replaced with direct DOM manipulation
 
 import { assert } from 'chai'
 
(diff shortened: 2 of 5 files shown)


phillipc:
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)
> ℹ️ Review info

Done.

brianmhunt (approved):
This is well-implemented and ready to merge. The three new callbacks fill a real gap that's bitten users of the reference build for a long time. A few specific things I verified:

- `computeMoves` correctly excludes primitives (no false-move callbacks when sorting string/number arrays) and uses `list.pop()` to match the LIFO reuse order of `pendingDeletes`, so duplicate-value arrays get consistent pairing.
- The `processQueue` `try/finally` now covers the full flush sequence — a throwing user callback won't wedge the binding with `rendering_queued` stuck at `true`.
- `afterRender` fires for both fresh template clones and recycled nodesets from `pendingDeletes`.
- Test coverage is thorough: initial render, push, reverse, multi-root templates, primitive arrays (no spurious move callbacks), and error-recovery paths are all exercised.
- All CI checks green across Chromium, Firefox, WebKit, and happy-dom.

Approving. Thanks for the careful iteration on the CodeRabbit feedback.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #373, opened 2026-04-27
plan: collapse release to a single human action

brianmhunt:
## Summary

Captures the cost of TKO's current tag-driven release flow — which we just lived through on the 4.1.0 cut — and proposes collapsing it to a **single human action: merge the version PR**.

## Why now

The 4.1.0 cut required:
1. Push initial tag `v4.1.0`
2. Wait for `changesets/action` (which silently skipped PR creation under the tag-push trigger — known wrinkle, branch name `changeset-release/refs/tags/v4.1.0` is the symptom)
3. Hand-open the version PR (#372) wrapping the bot-generated branch
4. Review + merge the PR
5. Force-move the tag to the merged commit and re-push
6. Wait for second workflow run

Five out of six steps are admin or latency. The two with maintainer judgment (review and merge) are buried in the noise. That cuts directly against the Dark Factory thesis — a single time-constrained maintainer with AI agents shouldn't be pushing tags twice and worrying about whether they're on the right commit.

## Proposed shape

- Trigger `release.yml` on `push: branches: [main]` (the canonical changesets pattern).
- Single job replaces three (`prepare-release` / `publish` / `github-release`).
- `changesets/action`'s built-in `publish` input runs `npx changeset publish` automatically when there are no pending changesets and sets `outputs.published`.
- A post-publish step generates `vX.Y.Z` from `tools/release-version.cjs`, creates the tag and GitHub Release. No more force-push tag dance, no more tag-vs-version validation.

## Tradeoffs

- Lose the "tag when I want to release" entry point. Maintainer instead decides timing by **when they merge the version PR**.
- Releases batch by version-PR cadence — every changeset-bearing feature PR rebumps the open version PR; merge whenever the batch feels release-worthy.
- GitHub release notes lose the "tag pre-existed" guarantee. Mitigation: `github-release.yml` already exists as a manual backfill workflow.

## Out of scope

- Changelog reform (separate plan — about artifact shape, not pipeline).
- Pre-release / canary channels (separate plan — would auto-publish every main commit to `next`).

## Test plan

- [ ] Plan reviewed for accuracy against current `release.yml` and the 4.1.0-cut experience
- [ ] Phasing approved (land plan + workflow rewrite together; test with a no-op patch release)

https://claude.ai/code/session_01Jnv99bwWtFWFmsyL7A2cJv

---
_Generated by [Claude Code](https://claude.ai/code/session_01Jnv99bwWtFWFmsyL7A2cJv)_

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive release process documentation outlining updated workflow structure, permissions configuration, concurrency handling, and verification procedures.

### DIFF
--- a/plans/single-action-release.md
+++ b/plans/single-action-release.md
@@ -1,0 +1,223 @@
+# Plan: Single-Action Release
+
+**Goal**: Reduce the TKO release flow from a multi-step manual dance (tag,
+wait, hand-open PR, merge, force-tag, wait) to a single human action — merge
+the auto-generated version PR. Everything else runs unattended.
+
+This is a Dark Factory plan: the maintainer's job is direction, not
+choreography.
+
+---
+
+## Original State (April 2026)
+
+`release.yml` triggers on `push: tags: v*`. The intended flow is:
+
+1. Maintainer pushes `vX.Y.Z` tag.
+2. `prepare-release` runs `changesets/action@v1`. If there are pending
+   changesets, it commits version bumps to a branch and opens a "chore:
+   version packages" PR.
+3. Maintainer reviews and merges that PR.
+4. Maintainer force-moves the tag to the merged commit and re-pushes.
+5. `prepare-release` re-runs, sees no changesets, gates in the `publish`
+   job. `changeset publish` runs to npm via OIDC.
+6. `github-release` creates the matching GitHub Release.
+
+### What actually happened on the 4.1.0 cut
+
+- Step 2 produced the bumped branch `changeset-release/refs/tags/v4.1.0`
+  but **did not open a PR**. `changesets/action@v1` was designed for
+  `push: branches` triggers; under tag-push it can produce the branch and
+  silently skip PR creation. The branch name itself leaks the tag ref
+  (`changeset-release/refs/tags/v4.1.0` instead of the usual
+  `changeset-release/main`), which is the symptom.
+- A human had to hand-open the PR (#372) wrapping the bot-generated
+  branch.
+- Step 4's force-push tag is a footgun — easy to forget to `git pull`
+  first and tag the wrong commit.
+
+## Pain points
+
+| Current step | Time/risk cost |
+|---|---|
+| Push initial tag | Trivial, but obscures intent ("am I starting a release or testing CI?") |
+| Wait for action to maybe-open a PR | Action is flaky under tag triggers (see above) |
+| Hand-open the PR if action skipped | Pure admin |
+| Review version bumps | Necessary, mechanical |
+| Merge PR | Necessary |
+| Force-move tag, re-push | Footgun: must be on merge commit |
+| Wait for second workflow run | Pure latency |
+
+Five out of seven steps are admin or latency. The two with maintainer
+judgment (review and merge) are buried.
+
+---
+
+## Target State
+
+**Trigger**: `push: branches: [main]`.
+
+**Flow**:
+1. Feature PR with a changeset merges to `main`.
+2. `release.yml` runs:
+   - Pending changesets exist → `changesets/action` opens or updates the
+     "chore: version packages" PR. Done.
+3. Maintainer accumulates PRs over time. When ready to release, **merges
+   the version PR**.
+4. `release.yml` re-runs:
+   - No pending changesets → `publish` job runs. `changeset publish`
+     bumps tags per-package on npm via OIDC.
+   - A post-publish step creates a single repo-wide `vX.Y.Z` tag and a
+     matching GitHub Release.
+
+**Single human action: merge the version PR.** Everything before and
+after is automation.
+
+---
+
+## Mechanism
+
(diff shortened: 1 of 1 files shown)


brianmhunt:
## Adversarial review

Verified against `.github/workflows/release.yml`, `tools/release-version.cjs`, and `AGENTS.md` § Release Process on this branch.

### Verified

- ✅ Current trigger: `push: tags: ['v*']`
- ✅ Three jobs: `prepare-release` / `publish` / `github-release`
- ✅ `tools/release-version.cjs` exists and is the version source
- ✅ `github-release.yml` exists as manual backfill
- ✅ `publish-check.yml` exists for PR validation
- ✅ AGENTS.md § Release Process documents the tag-push entry point and force-move dance (lines 156–159)
- ✅ `changesets/action`'s built-in `publish:` input runs `npx changeset publish` and exposes `outputs.published`

### Issues with the proposed `release.yml`

1. **Missing `concurrency:` block.** Current file has `concurrency: ${{ github.workflow }}-${{ github.ref }}`. Without it, two main pushes in quick succession (back-to-back PR merges) can race the version-PR update or the publish path. Keep it.

2. **`persist-credentials: false` + `git push origin \"$tag\"` won't work.** The proposed checkout disables credential persistence, but the post-publish step then runs `git tag` + `git push origin \"$tag\"`. With no credentials persisted, the push has no auth. Either set `persist-credentials: true` for that checkout (least bad) or create the tag via `gh api repos/:owner/:repo/git/refs -f ref=refs/tags/$tag -f sha=$GITHUB_SHA` using `GH_TOKEN`. Current `github-release.yml` sidesteps this by only creating the GH release (which gh CLI handles), not pushing a git tag.

3. **`commitMode: github-api` silently dropped.** Current `prepare-release` step uses `commitMode: github-api`, which makes `changesets/action` create the version-PR commit through the GitHub API (signed by github-actions[bot] / verified). Plan's snippet omits it — falls back to local `git commit`, producing unsigned commits and potentially tripping branch-protection rules that require verification. If intentional, say why; if not, keep it.

4. **Loss of least-privilege isolation.** Current splits `prepare-release` (`contents: write`, `pull-requests: write`) from `publish` (`contents: read`, `id-token: write`) so the publish path can't modify branches/PRs and the PR-opening path can't mint OIDC tokens. The proposed single-job collapses all three permissions onto every run. Worth at least a tradeoff bullet acknowledging this — or two-step the workflow (`changesets/action` job → conditional `publish-and-tag` job gated on `outputs.published`).

5. **`bun run build` runs on every main push.** Single-job design means the build runs even when the path is \"open/update version PR, no publish\". Current `prepare-release` skips the build for that path. On a busy main branch this is real CI minutes. Either accept it explicitly or gate the build behind `if: steps.changesets.outputs.published == 'true'` and run it before the publish step (note: `changesets/action`'s built-in `publish` runs *inside* the action, so build must precede the action step — push the build into a separate downstream job triggered by `outputs.published`, or live with the waste).

6. **Tag-creation ordering.** Plan does `git tag` → `git push origin \"$tag\"` → `gh release create`. If `gh release create` fails after the tag is pushed, the repo has an orphan tag with no release, and re-running won't help (the existing-tag check in `github-release.yml` is keyed on the *release*, not the tag). Safer ordering: `gh release create --target \"$GITHUB_SHA\" \"$tag\"` first (which creates the tag *and* release atomically when the tag doesn't yet exist), skip the explicit `git tag` + `git push`. `gh release create` will create the underlying ref for you.

7. **Auto-publish on next main push after merging the version PR is fine, but flag the implicit double-merge case.** Sequence: version PR auto-updated with one changeset → maintainer merges it → before publish completes, another feature PR with a new changeset lands on main → `release.yml` re-runs, sees no pending changesets at the moment it checked (race), publishes prematurely. Low probability but worth a sentence in *Tradeoffs*. The `concurrency:` block (issue 1) mostly mitigates.

### Document-level nits

- *Pain points* table has 7 rows; *Original State* numbered list has 6 steps. \"Five of seven\" vs \"Five of six\" appear in adjacent paragraphs — the table inserts an implicit \"review version bumps\" row that the numbered list folds into step 3. Reconcile.
- Phasing step 1 says \"land plan + workflow rewrite together\" but step 2 defers AGENTS.md update. AGENTS.md § Release Process actively misleads as soon as the workflow flips — it should change in the same PR or the plan should call out the temporary inconsistency.
- *Out of scope* mentions canary channels but the post-publish snippet already branches on `*-alpha*|*-beta*|*-rc*` for `--prerelease`. Worth one line explaining what the prerelease branching *does* cover (git/GH-release marking) vs what's still out of scope (auto-publishing every main commit to `next`).

### Verdict

Direction is right and the dark-factory motivation is well-grounded. Request changes for issues 1–4 (concurrency, push-auth, commitMode, least-privilege) before this becomes the implementation; 5–7 are tradeoffs to acknowledge in the plan rather than blockers.

Issue #11, opened 2016-11-22
Cannot find closing comment tag to match:  ko with: blog


jlaustill:
[Edit.tmpl.html.tar.gz](https://github.com/knockout/tko/files/607092/Edit.tmpl.html.tar.gz)
After fixing all my click bindings, I'm now getting this error.  I'm attaching the view.  This works in 3.4.0 and could very well be another thing I'm doing wrong and stupid, or a bug in tko.  I went over the view and nothing popped out at my, maybe it will at you.  Error and stack below, view attached.

```
ko.js:898 Uncaught Error: Cannot find closing comment tag to match:  ko with: blog 
    at getVirtualChildren (http://local.north40net.com/node_modules/tko/dist/ko.js:898:17)
    at getMatchingEndComment (http://local.north40net.com/node_modules/tko/dist/ko.js:903:32)
    at nextSibling (http://local.north40net.com/node_modules/tko/dist/ko.js:1013:18)
    at invokeForEachNodeInContinuousRange (http://local.north40net.com/node_modules/tko/dist/ko.js:6236:25)
    at activateBindingsOnContinuousNodeArray (http://local.north40net.com/node_modules/tko/dist/ko.js:6256:15)
    at executeTemplate (http://local.north40net.com/node_modules/tko/dist/ko.js:6331:11)
    at computed.disposeWhen (http://local.north40net.com/node_modules/tko/dist/ko.js:6373:44)
    at Function.evaluateImmediate_CallReadThenEndDependencyDetection (http://local.north40net.com/node_modules/tko/dist/ko.js:2667:105)
    at Function.evaluateImmediate_CallReadWithDependencyDetection (http://local.north40net.com/node_modules/tko/dist/ko.js:2639:31)
    at Function.evaluateImmediate (http://local.north40net.com/node_modules/tko/dist/ko.js:2603:20)
[...]

PR #374, opened 2026-04-27
plan: single structured root CHANGELOG (B + C)

brianmhunt:
## Summary

The default `changeset version` writes a `CHANGELOG.md` per package. With TKO's 27-package fixed-version group, a single cross-cutting concern lands in eight package files verbatim — `modernize-utils-dead-polyfills` did exactly this on the 4.1.0 cut. To answer *"what landed in 4.1.0?"* an agent has to find every CHANGELOG, dedupe, and reconstruct a story we already wrote once in `.changeset/*.md`.

This plan replaces that with **a single root `CHANGELOG.md` carrying structured per-release entries** (B + C combined).

## Options surveyed

- **A.** Top-level CHANGELOG, per-package stubs.
- **B.** Single source, no per-package files.
- **C.** B + structured release notes (curated summary + bullet highlights + package list + changeset references).

## Recommendation: B + C together

The same custom changelog generator that makes B work is the hook C needs for structured output — doing them in one PR is one piece of work, not two. The maintainer (or agent acting as release manager) edits a placeholder summary before merging the version PR; for trivial patch releases, the placeholder ships as-is.

## Implementation surface

- `tools/changesets/root-changelog.cjs` — implements the Changesets `getReleaseLine`/`getDependencyReleaseLine` interface.
- `tools/changesets/version.cjs` — wraps `changeset version`, deletes per-package CHANGELOGs, aggregates into one structured root entry.
- `.changeset/config.json` — points at the custom generator.
- `release.yml` — one-line change to `version: node tools/changesets/version.cjs`.
- One-time cleanup: `rm packages/*/CHANGELOG.md builds/*/CHANGELOG.md`.

## Tradeoffs

- npm package pages no longer show per-package histories — point them at the repo CHANGELOG via package descriptions.
- Authoring discipline: maintainer edits placeholder summary for non-trivial releases.
- Renovate/Dependabot fall back to GitHub Release when CHANGELOG.md is absent — verify before landing.

## Out of scope

- Pipeline reforms (covered in `plans/single-action-release.md` / PR #373).
- Per-release blog posts and migration guides (live in `tko.io/`, not CHANGELOG).

## Test plan

- [ ] Plan reviewed for accuracy
- [ ] Recommendation (B+C together vs B-then-C) approved
- [ ] Implementation surface scoped (tools/, config, workflow change, cleanup)

https://claude.ai/code/session_01Jnv99bwWtFWFmsyL7A2cJv

---
_Generated by [Claude Code](https://claude.ai/code/session_01Jnv99bwWtFWFmsyL7A2cJv)_

### DIFF
--- a/plans/changelog-dark-factory.md
+++ b/plans/changelog-dark-factory.md
@@ -1,0 +1,271 @@
+# Plan: Dark-Factory Changelog
+
+**Goal**: Replace TKO's per-package changelog sprawl with a single
+canonical release narrative that agents and humans can grep once.
+
+This is a Dark Factory plan: the artifact shape should optimize for
+*token-efficient retrieval by an agent*, not for the
+one-package-one-history mental model inherited from Lerna-era npm
+tooling.
+
+---
+
+## Original State (April 2026)
+
+`changeset version` writes a `CHANGELOG.md` per package. With TKO's
+fixed-version group of 27 packages, a single cross-cutting concern
+(say, `modernize-utils-dead-polyfills`) appears in **eight** package
+changelogs verbatim — because the changeset listed all eight packages
+that touched the work. Sister packages dragged along by the
+fixed-group bump get a single-line `Updated dependencies` stub.
+
+### Concretely on 4.1.0
+
+| Changeset                          | Full body appears in N CHANGELOGs |
+|------------------------------------|-----------------------------------|
+| `modernize-utils-dead-polyfills`   | 8                                 |
+| `modernize-trigger-event`          | 3                                 |
+| `fix-jsx-clean-teardown-race`      | 1                                 |
+| `fix-proxy-delete-property`        | 1                                 |
+
+To answer *"what landed in 4.1.0?"* an agent (or human) currently has
+to:
+
+1. Find every `CHANGELOG.md` in the monorepo.
+2. Read each file's `## 4.1.0` section.
+3. Dedupe the bodies.
+4. Reconstruct the cross-cutting narrative.
+
+That's four wasted operations to recover a story we already wrote
+once, in `.changeset/*.md`.
+
+## Pain points
+
+| Symptom                                                    | Cost                                          |
+|------------------------------------------------------------|-----------------------------------------------|
+| Same changeset body duplicated across 8 files              | Storage trivial; cognitive cost real          |
+| No single answer to "what's in this release"               | Agents must aggregate; humans must scroll     |
+| Release notes on GitHub redundantly auto-generate the same | Three sources of truth (PR, CHANGELOG, release) |
+| Per-package `## 4.0.0` history accumulates noise           | Fixed-group means every package sees every release |
+
+The default Changesets layout assumes Lerna-style independent
+packages, each with its own audience reading its own changelog. TKO
+ships as a fixed group with one version line — the model doesn't fit.
+
+---
+
+## Options
+
+### A. Top-level CHANGELOG, per-package stubs
+
+- Custom `getReleaseLine`/`getDependencyReleaseLine` writes the
+  narrative once to root `CHANGELOG.md`.
+- Per-package `CHANGELOG.md` shrinks to:
+  ```
+  ## 4.1.0 — see [root CHANGELOG](../../CHANGELOG.md#410)
+  ```
+- npm consumer pages still render *something* per package.
+- Smallest behavioral change; each release adds two files modified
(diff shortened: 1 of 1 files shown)


brianmhunt:
## Adversarial review

Verified against repo state on `main`.

### Verified ✅

- 8 CHANGELOG.md files contain the `modernize-utils-dead-polyfills` body verbatim. Pain point is real.
- `getReleaseLine` / `getDependencyReleaseLine` are the correct interface (per `@changesets/types`).
- All 27 packages' `package.json` `files` arrays are `["dist/"]` — CHANGELOG.md is not included in published tarballs.

### Findings

1. **Package count is 27, not 26.** `.changeset/config.json` `fixed[0]` has 27 entries (verified via `jq '.fixed[0] | length'`). One adversarial pass got this wrong; just confirming.

2. **[blocker] `version.cjs` second-run is racy.** The plan says step 2 is "read what would have been written, then delete". `changeset version` *appends* to existing `CHANGELOG.md`. After the one-time cleanup, file does not exist for run #1 — clean. Run #2: file gets created by `changeset version` again. Run #3+: the wrapper would be reading multi-release files and needs a rule to distinguish "just-written" from "history". Plan does not specify. Either the wrapper deletes after every run (seems intended but not stated) or it diff-detects the new section.

3. **[blocker] Conflicts with PR #377 / plan #373.** PR #377 (already open against main) sets `version: npx changeset version` in `release.yml`. This plan changes it to `version: node tools/changesets/version.cjs`. The two PRs edit the same field; whichever lands second silently overrides the first. Order must be defined: this plan's impl PR rebases on #377, or the field is left as `npx changeset version` until both plans converge in a single impl PR.

4. **[blocker] Verification command is broken.** Plan: `find . -name CHANGELOG.md -not -path './node_modules/*'`. The repo also has `.claude/worktrees/*/packages/*/CHANGELOG.md` and `.claude/worktrees/*/builds/*/CHANGELOG.md` (Claude worktree clones). The check needs to exclude `.claude/worktrees` and `.git/worktrees` or it returns false negatives forever.

5. **[tradeoff] npm-tarball motivation is softer than stated.** The CHANGELOG.md is *already* excluded from every published tarball (`files: ["dist/"]` in all 27). npm package pages don't render CHANGELOG content from the tarball — they render it from the repo URL (`repository.url` field). Removing per-package CHANGELOG.md *does* break the "open npm page → click CHANGELOG link" path, but it doesn't change tarball content. The plan's "npm pages no longer show per-package histories" framing implies the tarball matters; it doesn't. Re-state the cost as "deep-link experience for downstream consumers", not "tarball content".

6. **[tradeoff] "nx, pnpm itself" comparison cuts the wrong way.** Both `nx` and `pnpm` *do* ship per-package CHANGELOG.md files. Spot-check before relying on this in the rationale.

7. **[tradeoff] Renovate fallback is unverified.** Plan says "Renovate falls back gracefully to the GitHub Release when CHANGELOG.md is absent" but lists this in *Tradeoffs* as "verify before landing". Recommendation has already committed to B+C while the listed risk is open. Renovate's `release-notes` extractor reads `CHANGELOG.md` from the repo and renders nothing if absent — it does *not* automatically swap to the GitHub Release. Verify before recommendation hardens.

8. **[tradeoff] Deleted-history is asymmetric with the rest of the plan.** Mitigation = "git log". External consumers who deep-linked to `packages/utils/CHANGELOG.md#400` will 404. Suggest at least a redirect note in the release announcement, or a `CHANGELOG-archive.md` at the package root with a one-line "see root CHANGELOG for current". Cheap, preserves discoverability.

9. **[nit] Structured template hard-codes "All N fixed-group packages → `X.Y.Z`".** True today (fixed group covers everything published). If a package ever leaves the group (e.g. a builds/* goes independent), the template silently lies. Generate the count from `.changeset/config.json` at version time instead of hard-coding.

10. **[nit] `.changeset/config.json` snippet uses `null` as the second array element** (changelog config). Valid per changesets schema but unusual — most generators take an options object. Plan should call out *why* `null` here (no per-changeset config; all formatting decisions are in the generator).

11. **[nit] Footnote `[1]` at the bottom is the only inline-link reference in that style.** Inconsistent with the rest of the doc.

### Verdict

Direction is right and the dark-factory framing fits TKO's reality (fixed group, agent-readable artifact). Request changes for: blockers #2, #3, #4 before this becomes the implementation; #5–#8 reframe in the doc; #9–#11 nits. Coordinate landing order with #377.

PR #376, opened 2026-05-02
fix: resolve async completion in ConditionalBindingHandler

phillipc:
## Summary

Fixes two bugs in `ConditionalBindingHandler` identified in [typescript-code-review-findings-5.md](https://github.com/Auge19/tko/blob/pc/review_skill_and_demo_plans/plans/typescript-code-review-findings-5.md) (Critical Finding 1).

- **Non-rendering branches never resolved**: `if: false`, `ifnot: true`, and `with: null` all hit the empty-branch path in `render()` without calling `completeBinding()`, leaving the `AsyncBindingHandler` promise permanently unresolved. Any `await applyBindings(...)` would hang indefinitely.

- **Premature completion with async descendants**: `renderAndApplyBindings` used `await applyBindingsToDescendants(...)`, but that function returns a `BindingResult` synchronously — not a Promise. The `await` resolved immediately without waiting for async child bindings to finish. Fix follows the `DescendantBindingHandler` pattern: call without `await`, use `bound.completionPromise` when `!bound.isSync`, and short-circuit `this.bindingCompletion = bound` for the sync case so the binding isn't unnecessarily tracked as async.

## Test plan

- [x] `bun run build` passes
- [x] `bunx vitest run packages/binding.if` — new regression tests for `if: false`, `ifnot: true`, `with: null` completion
- [x] `bunx vitest run packages/bind/spec/bindingCompletionPromiseBehavior.ts` — new test for `if: true` with async descendants
- [x] `bunx vitest run` — full suite (290 files, no regressions)

### DIFF
--- a/packages/binding.if/src/ConditionalBindingHandler.ts
+++ b/packages/binding.if/src/ConditionalBindingHandler.ts
@@ -76,6 +76,7 @@ export default class ConditionalBindingHandler extends AsyncBindingHandler {
       this.renderAndApplyBindings(this.ifElseNodes.elseNodes)
     } else {
       virtualElements.emptyNode(this.$element)
+      this.completeBinding()
     }
   }
 
@@ -83,7 +84,12 @@ export default class ConditionalBindingHandler extends AsyncBindingHandler {
     if (!useOriginalNodes) {
       virtualElements.setDomNodeChildren(this.$element, cloneNodes(nodes))
     }
-    const bound = await applyBindingsToDescendants(this.bindingContext, this.$element)
+    const bound = applyBindingsToDescendants(this.bindingContext, this.$element)
+    if (bound.isSync) {
+      this.bindingCompletion = bound
+    } else {
+      await bound.completionPromise
+    }
     this.completeBinding(bound)
   }
 
--- a/.changeset/fix-conditional-binding-async-completion.md
+++ b/.changeset/fix-conditional-binding-async-completion.md
@@ -1,0 +1,5 @@
+---
+"@tko/binding.if": patch
+---
+
+Fix async completion in ConditionalBindingHandler: call completeBinding() for non-rendering branches (if: false, ifnot: true, with: null) and correctly await async descendants via completionPromise instead of awaiting the BindingResult directly.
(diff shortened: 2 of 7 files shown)


brianmhunt:
## Adversarial review

Fix is faithful to the canonical `DescendantBindingHandler` pattern. Both bugs verified: `applyBindings` outer Promise hangs pre-fix on no-render branches (handler's `bindingCompletion` never resolves), and the bogus `await applyBindingsToDescendants(...)` resolved before async descendants completed. All four added tests would have failed pre-fix.

**Worth verifying while here:**
- Coverage gap: no test for `if: false` chained with `else` sibling (`else` binding chain). Same code path benefits from the fix; one extra spec line confirms it doesn't regress.

**Out of scope (notes for the record):**
- ConditionalBindingHandler now contains a near-verbatim copy of `DescendantBindingHandler.applyBindingsToDescendants` logic. Worth a refactor pass later — not blocking this PR.
- Changeset target (`@tko/binding.if` patch) is correct. Timing change is a bug fix, not a breaking contract.

LGTM modulo the optional else-chain test.

phillipc:
@brianmhunt: I have add a new test case to close the gap, but the previous was bigger ;)

packages/binding.if/src/ConditionalBindingHandler.ts
```diff
@@ -76,14 +76,20 @@ export default class ConditionalBindingHandler extends AsyncBindingHandler {
...
   }
 
   async renderAndApplyBindings(nodes: ArrayLike<Node>, useOriginalNodes?: boolean) {
     if (!useOriginalNodes) {
       virtualElements.setDomNodeChildren(this.$element, cloneNodes(nodes))
     }
-    const bound = await applyBindingsToDescendants(this.bindingContext, this.$element)
+    const bound = applyBindingsToDescendants(this.bindingContext, this.$element)
+    if (bound.isSync) {
+      this.bindingCompletion = bound
```
brianmhunt, line 89: Is this used for anything?  Seems new, but unused.  

Would this be sufficient?

```ts
if (!bound.isSync) {
  await bound.completionPromise
}
```

brianmhunt (commented):
Quick question/check, but looking good.  Ty @phillipc !

packages/binding.if/src/ConditionalBindingHandler.ts
phillipc: L.93 must be called with bound

phillipc: See https://github.com/knockout/tko/blob/main/packages%2Fbind%2Fsrc%2FBindingHandler.ts#L124-L126

brianmhunt (approved):
👍 looks good, ty @phillipc

PR #377, opened 2026-05-04
chore(release): collapse release flow to single human action

brianmhunt:
## Summary

Implements the plan in #373 — switch `release.yml` from tag-push trigger to push-to-main trigger, collapse three jobs into two, and make merging the version PR the only human action in a release.

### `.github/workflows/release.yml`

- **Trigger** flips from `push: tags: v*` to `push: branches: [main]`.
- **Two jobs** (`prepare` / `publish-and-tag`) preserve least-privilege isolation:
  - `prepare` holds `contents:write` + `pull-requests:write` (no OIDC).
  - `publish-and-tag` holds `contents:write` + `id-token:write` (no PR write).
- **`prepare`** uses `persist-credentials: false` paired with `commitMode: github-api`, so `changesets/action` commits the version bump via the API (verified, `github-actions[bot]`-authored).
- **`publish-and-tag`** is gated on `hasChangesets == 'false'`, so feature PRs with changesets do not trigger the publish path — their changesets accumulate in the version PR.
- **Build + test** run inside `publish-and-tag`, gated by the outer job condition, so no-changeset paths skip them.
- **Publish** uses `changesets/action`'s `publish` input. Its `published` output is the authoritative "did anything go to npm" signal — closing the trap where a plan-only / doc-only main push would otherwise hit `gh release create` for an already-existing tag.
- **Per-package git tags** suppressed via `--no-git-tag`; the repo-wide `vX.Y.Z` is the single tag.
- **`gh release create`** creates tag ref + release in one API call. No more `git tag && git push` orphan-tag failure mode.
- **Concurrency block** (`workflow-ref`) serializes against itself so two near-simultaneous main pushes can't race.
- **Prerelease detection** anchored on canonical suffixes (`*-alpha`, `*-alpha.*`, etc.) — no false positives like `1.0.0-alphabet`.

### `AGENTS.md`

- § Release Process rewritten around the single-action flow. Drops force-push-tag instructions.
- Workflows table row for `release.yml` updated (trigger column: "Push to main").

## Test plan

- [ ] **Land #373 first** (the plan PR) so `plans/single-action-release.md` is on `main` when this PR merges and the workflow comment chain stays consistent. (Plan PR was deliberately not bundled here so the design discussion stayed on the plan.)
- [ ] **No-op patch release** as the first run after this lands: one trivial changeset (e.g. typo), watch the version PR appear, merge it, watch publish run, confirm `vX.Y.Z` tag + GH release on the version-PR-merge commit.
- [ ] Verify a doc-only push to `main` (e.g. README edit) runs only `prepare`, exits with `should_publish=true`, runs `publish-and-tag`, sees `outputs.published=false`, skips the tag step. (This is the "wasteful but safe" path described in the plan's *Doc-only main pushes* tradeoff.)
- [ ] Verify a feature-PR-with-changeset merge runs only `prepare`, exits with `should_publish=false`, skips `publish-and-tag` entirely, and updates the open version PR.

## Notes

This PR depends on #373 conceptually but does not require it to merge first to function — the workflow stands on its own. Recommend landing #373 first for review-trail clarity.

## Adversarial review

Performed in commit message of `78d156f9`. Fresh subagent flagged the gating-logic bug (conflating "no pending changesets" with "version PR just merged"); fixed by switching to `changesets/action`'s `published` output. Remaining items accepted as documented tradeoffs.

## Summary by CodeRabbit

* **Chores**
  * Modularized CI by extracting build-and-test into a reusable workflow and updated main pipeline to delegate to it.
  * Simplified release pipeline to a single prepare flow that gates publish steps and runs the reusable build-and-test.
  * Added workflow concurrency control to avoid overlapping runs.

* **Documentation**
  * Updated release process docs to reflect the new automated release flow and maintainers' guidance.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -1,0 +1,54 @@
+# Reusable workflow: build + browser-matrix test on a checked-out tree.
+#
+# Called by:
+#   - main-build.yml (push to main, workflow_dispatch)
+#   - release.yml    (publish gate)
+#
+# Single source of truth for "build + matrix test". Other workflows
+# express dependency via `needs:` on the calling job, so cross-workflow
+# coordination uses standard GitHub Actions semantics — no polling.
+name: Build and Test
+
+on:
+  workflow_call:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    container:
+      image: mcr.microsoft.com/playwright:v1.59.1-noble
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+
+      # Install Bun directly — avoids `apt-get update` (9+ min in this
+      # container) which setup-bun needs for unzip.
+      - name: Install Bun
+        run: python3 tools/install-bun
+
+      - name: Install dependencies
+        run: bun install --frozen-lockfile
+
+      - name: Run Build
+        run: bun run build
+
+      - name: Verify ESM extensions
+        run: bun run verify:esm
+
+      - name: Run tests (browser matrix)
+        run: bunx vitest run --project browser
+        env:
+          HOME: /root
+          VITEST_BROWSERS: chromium,firefox,webkit
+
+      - name: Run tests (cli-happy-dom)
+        run: bunx vitest run --project cli-happy-dom
+        env:
+          HOME: /root
+
+      - name: Upload build artifacts
+        uses: actions/upload-artifact@v7
+        with:
+          name: tko
+          path: builds/knockout/dist
--- a/.github/workflows/main-build.yml
+++ b/.github/workflows/main-build.yml
@@ -1,4 +1,6 @@
-# workflow for build and deploy
+# Push-to-main + manual-dispatch entry point for the build-and-test
+# reusable workflow. Real logic lives in build-and-test.yml so
+# release.yml can depend on it via `needs:`.
 name: Build
 
 on:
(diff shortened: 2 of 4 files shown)


brianmhunt:
## Addressed in 29d8dec8

**Codex P1 (blocker):** dropped `--no-git-tag`. Confirmed `changesets/action` parses `New tag:` lines from `changeset publish` stdout to set `outputs.published`; suppressing those lines made the gate dead. Per-package tags now ship (27 per release) — noise but harmless. Canonical release marker remains the repo-wide `vX.Y.Z` tag created by `gh release create`.

**Copilot:**
- ✅ `npx` → `bunx` for `changeset version` and `changeset publish`. Matches AGENTS.md tooling.
- ✅ Header comment reworded — publish-and-tag runs on any no-pending-changeset main push (docs/plans included), not only after version-PR merge. The `published` gate is what makes the path safe.
- ✅ persist-credentials note sharpened: `gh release create` uses GH_TOKEN; the credential persistence is for `changeset publish`'s per-package tag pushes via git remote.

## Not addressed (by design)

- **Double `bun run build`** (Copilot): yes — once in `build-and-test` (reusable, in Playwright container, gates publish via `needs:`) and once in `publish-and-tag` (in the runner that will execute `changeset publish`). Reusable workflows do not share artifacts unless `actions/upload-artifact` + `download-artifact` is wired through, and our existing artifact (`tko`, `builds/knockout/dist`) covers only one build's output, not all 27 packages' dists. Future optimization: expand the upload to cover every `packages/*/dist` and `builds/*/dist`, then download in publish-and-tag and skip the local build. Out of scope for this PR.

- **`bun run test` step inheriting `VITEST_BROWSERS=chromium`**: the step was already removed when `build-and-test` became the gate (commit `7aefbc45`). The reusable workflow runs the full browser matrix in the Playwright container, and `publish-and-tag` `needs: build-and-test`. No tests run locally in `publish-and-tag` anymore.

PR #364, opened 2026-04-22
feat: add test coverage run with vitest + v8

phillipc:
## Summary

Recovering a lost feature: Adds a `bun run test:coverage` script that produces a v8 coverage report against the authoritative chromium browser project.

## Changes

- **`package.json`** — new `test:coverage` script: `bun run build && bunx vitest run --coverage --project browser`. Adds `@vitest/coverage-v8` as a devDependency (pinned exact `4.1.4` — its declared `peerDependencies: { vitest: 4.1.4 }` is exact, so a caret on both sides would warn on the next patch bump).
- **`vitest.config.ts`** — adds a top-level `coverage` block (v8 provider; reporters: `text`, `text-summary`, `html`, `lcov`, `json-summary`; outputs to `coverage/`). Includes `packages/*/dist/**/*.js` (so v8 picks up actually-executed code) and both `packages/*/src/**/*.ts` and `builds/*/src/**/*.ts` (so source-map remapping surfaces TS line numbers in the report). Index-barrel exclude is anchored to `packages/` only so `builds/{knockout,reference}/src/index.ts` (the only real src file in each build) is not silently dropped.
- **`tools/build.ts`** — switches esbuild from `--sourcemap=external` to `--sourcemap=linked`. Published `dist/*.js` gain one trailing `//# sourceMappingURL=...` comment line so v8 can remap coverage back to the original TS sources. `.map` files were already shipped in npm tarballs (`files: ["dist/"]`), so no consumer-facing 404 risk.
- **`.gitignore`** — adds `COVERAGE.md`. The snapshot rots on every run that isn't followed by a commit; reintroduce later when `test:coverage` deterministically rewrites it.

## Verification

- `bun run test:coverage` exits 0; report written to `coverage/`.
- The authoritative real-browser matrix (`bun run test`) is unchanged — coverage is opt-in via the new script.

## Notes

- Coverage runs against the chromium project only. Firefox/WebKit are not covered — JavaScriptCore and SpiderMonkey don't expose V8's coverage protocol. The matrix's job is correctness across engines; coverage's job is "what code did we exercise" — Chromium is a sufficient proxy because the spec set is identical across all three browsers.
- Published `dist/*.js` artifacts gain one trailing `//# sourceMappingURL=...` comment line. No other build-output changes.
- Coverage is not gated in CI; that's a follow-up. Land opt-in, gate later once stability is proven.

## New Features
* Introduced test coverage reporting with support for multiple output formats (HTML, LCOV, JSON, and text summaries).

[...]

PR #378, opened 2026-05-05
docs(AGENTS): mandate pre-merge verification

brianmhunt:
## Summary

Adds a new \`## Pre-Merge Verification\` section to AGENTS.md mandating three checks before declaring any PR with a runnable artifact ready:

1. **Run it.** Local invocation, exit 0, expected artifact. \`bun install\` to sync deps when verifying a PR is part of the work.
2. **Test it.** Existing test exercises the new path, or a new spec / smoke test added in the same PR.
3. **CI parallel.** \`gh run list --branch <branch>\` shows every triggered workflow GREEN.

## Why

Surfaced today on PR #364 (coverage harness): pushed without local run; first push lint-failed on a Biome formatting violation, then local invocation failed on a missing dep. Both 60-second pre-checks the agent should have done before declaring the PR ready. User had to prompt with "did we run it?" to surface gaps that should have been mandatory upfront.

The dark-factory thesis (small team + AI agents) only works if agents close the verification loop without prompting. Long PR threads where a maintainer has to chase basic checks defeat the throughput model.

## What changed

One new section in AGENTS.md, between "Always Improve" and "Review Your Own Change Adversarially". 10 lines of prose, explicit output convention ("ran locally: ✓; CI green: ✓; smoke test added: ✓").

## Test plan

- [ ] Section reads as standalone, doesn't duplicate adjacent guidance
- [ ] Output convention is mechanical enough to verify in PR review
- [ ] Future agents pick this up via the standard AGENTS.md load

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -218,6 +218,16 @@ When a feedback loop fails, fix the loop — not just the symptom. Examples: `bu
 
 Avoid scope creep. If an improvement would balloon the PR, file a follow-up issue or spawn a separate task instead.
 
+## Pre-Merge Verification
+
+A PR that adds or modifies a runnable artifact (script, workflow, CLI entry point, build target) is not mergeable until **all three** of these are done — without waiting to be asked:
+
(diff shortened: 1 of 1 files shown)


phillipc:
@brianmhunt I would like to share at this point some options to save tokens:

1) I think we need a instruction to build the solution before any validations. I often found ai-loops like this:
`I found why the tests still timed out: these specs import from packages/binding.if/dist, so they were still exercising stale built output. I’m rebuilding binding.if first, then rerunning the scoped tests with the new case.` _Update: We can also with to import from src._

2) The “Adversarial review” is good but expensive. We can probably make it risk-dependent or per release.

Issue #403, opened 2026-06-05
Add published-type checks (attw + publint) to publish-check CI


brianmhunt:
## Context

PR #402 makes TKO publish per-package `.d.ts` declarations for the first time. Once that lands, the published type/packaging surface becomes something we can regress silently — a future package could ship without a `types` entry, with a broken `exports` ordering, or with a declaration that doesn't resolve. We should guard it in CI.

This is a **follow-up to #402** — the checks validate the types that #402 introduces, so this only makes sense once #402 is merged.

## Proposal

Add two standard tools to the `publish-check.yml` workflow (which already runs on **every PR** to verify packages are publishable):

- **`publint`** — validates `package.json` packaging correctness (exports condition ordering, `files`, etc.). Resolution-mode-agnostic, near-zero false positives. Straightforward win.
- **`@arethetypeswrong/cli` (`attw --pack`)** — validates that published declarations actually resolve.

Optionally also wire both into the local `verify` script so they're caught before a PR is even pushed.

## Important: scope attw to TKO's actual audience

TKO is a browser MVVM framework; its consumers use bundlers (Vite/webpack/esbuild/Rollup → `moduleResolution: bundler`). The `node16`/`nodenext` TypeScript resolution mode is **out of audience** — a no-bundler Node+TS consumer is not a realistic TKO user.

Out of the box, attw fails on the node16/nodenext columns. We should add an `.attw.json` that **ignores the node16-specific rules** (e.g. `false-cjs`, `false-esm`, `false-export-default`, `unexpected-module-syntax`, `internal-resolution-error`, `cjs-resolves-to-esm` — exact names to be confirmed against `attw --list-rules` at implementation time) so the gate fails only on genuinely-relevant problems for bundler consumers (e.g. `no-resolution`, `untyped-resolution`). The ignored rules should be documented in `.attw.json` so the decision is explicit, not a silent suppression.

## Evidence — current attw matrix (run locally on #402's branch)

| Package | node10 | node16 (CJS) | node16 (ESM) | bundler |
|---|---|---|---|---|
| `@tko/observable` (CJS-default leaf) | 🟢 | 🟢 | 🚭 UnexpectedModuleSyntax | 🟢 |
| `@tko/utils` (`type:module`) | 🟢 | 👺 FalseESM + 🥴 InternalResolutionError | 🥴 InternalResolutionError | 🟢 |
[...]

PR #404, opened 2026-06-07
403 add published type checks attw publint to publish check ci

phillipc:
**Combination of #402 and #403** 

Implemented the issue as a full package-surface fix rather than an attw ignore. TKO now publishes mode-aware declarations and export maps so bundler, node16, and nodenext all resolve cleanly. The core changes are in tools/build-dts.ts, tools/verify-publish-types.ts, package.json, .github/workflows/publish-check.yml, plus the public package.json files under packages/* and builds/*. I also added the required plan at plans/2026-06-07-module-resolution-publish-check.md.

What changed: the declaration build now emits .d.mts and .d.cts siblings alongside .d.ts, rewrites relative specifiers to .mjs/.cjs as needed, and special-cases the package-level default-export entrypoints that need export = on the CommonJS side. Public manifests now use nested exports so import gets ESM JS plus .d.mts, and require gets CJS JS plus .d.cts. I also added publint and @arethetypeswrong/cli to local/CI verification, and expanded the type smoke tests with node16 and nodenext configs plus CJS/ESM consumers.
Verified locally with bun run dts, bun run build, bun run tsc, bun run verify:types, bun run verify:publish-types, and bun run verify:esm. attw --pack is now green across the full public package set. I did not run the full Vitest suite. I also tried bun run check; it still fails on pre-existing Biome findings in packages/utils.parser/src/*, but the new/changed file tools/build-dts.ts passes Biome on its own.

feat: enhance type verification and module resolution for TKO packages

- Updated package.json scripts to include additional type verification for Node16 and Nodenext configurations.
- Added new `verify:publish-types` script to ensure publishable packages meet type and output requirements.
- Modified package exports to support ESM and CJS formats with appropriate type declarations.
- Introduced new TypeScript configuration files for Node16 and Nodenext compatibility.
- Enhanced build-dts script to generate mode-aware declaration files.
- Created smoke tests for type verification in both CommonJS and ESM formats.
- Updated all package.json files to use .mjs for ESM entry points and .cts for CJS declarations.


tools/build-dts.ts
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env bun
+import { $, Glob } from 'bun'
+
+const tempOutDir = '.dts-tmp/emit'
+const buildWorkspaces = ['builds/knockout', 'builds/reference']
+const packageWorkspaces: string[] = []
+// These entrypoints publish a package-level default export, but their CommonJS
+// runtime shape is module.exports = value rather than { default: value }.
+const cjsEntrypointsNeedingExportEquals = new Set(['builds/knockout', 'builds/reference', 'packages/utils.component'])
```
phillipc, line 9: utils and utils.component can also be changed. But I think this is more a task for tko5 and the reference build focus.

@brianmhunt what do you think?

Issue #414, opened 2026-07-08
if binding evaluating to false delays invocation of koDescendantsComplete


brimimc:
Migrating from knockout, I have component templates that have simple if binding checks like 

`
`

if `showTabsList` is set to `false`, and we have a `koDescendantsComplete()` function, that function is never called. Only when `showTabsList` is set to true will `koDescendantsComplete` be called.

This appears to be similar behavior to knockout's `completeOn: "render"` option, but that's not what we want/need.

Is there any way to have `koDescendantsComplete` even when an `if` binding is bound to `false`?

brimimc:
This may be fixed in [latest](https://github.com/knockout/tko/commit/6f94126246988793a09239c1dde30a380cf20b74) - will need to get a release to test with

brimimc:
Of note - we've found that the descendantsComplete binding handler applies TOO EARLY (before a child foreach has updated the DOM) - I'm not sure if the comment on the fix "Fix mirrors DescendantBindingHandler" if that will also be too early.

@phillipc any ideas when the next release will happen for me to test the updates? thanks

phillipc:
Hey @brimimc, i think the `DescendantsComplete`-behavior is different in TKO. 

Properly you can switch to `childrenComplete: render`, see here: https://github.com/knockout/tko/blob/f3965180ebd2126bdeb22b8530b0876f84ba7b18/packages/binding.core/spec/descendantsCompleteBehaviors.ts#L118

BTW: Have you already set the option `createChildContextWithAs` to `false` for your migration project (see #191)?

brimimc:
Interesting @phillipc ... switching from `descendantsComplete` to `childrenComplete` could work for a single sub-component, but we have multiple and we need to wait for all of them to have `childrenComplete` before invoking our logic.

And yes we do have `ko.options.createChildContextWithAs = true;` set.

Any update on when the next release with koDescendantsComplete behavior fixes might be released?

phillipc:
I couldn't stop thinking about this either. I've implemented a solution that restores the KO behavior. 

@brianmhunt , please take a look at this. -> #416

PR #415, opened 2026-07-16
#414 - docs: descendantsComplete behavior and migration guidance for TKO

phillipc:
Documentation for a breaking change in TKO: `descendantsComplete` no longer waits for async descendants

See: #414 

## Summary by CodeRabbit

* **Documentation**
  * Updated the 3-to-4 migration guide with clearer `descendantsComplete` behavior details.
  * Explained the new per-node, synchronous timing and how async bindings affect completion.
  * Documented the edge case for nested `if`/`template` content that may render later without re-triggering.
  * Advised switching to `childrenComplete` when waiting for descendants is required, with updated example snippets.


phillipc:
@coderabbitai approve

